PR #44, opened 2023-03-30
[CSL-2238] Bundle

esezen:
Ways to use:

```js
window.ConstructorioQuizzes({
  selector: "#app",
  includeCSS: false,
  apiKey: "",
  quizId: "coffee-quiz",
  resultsPageOptions: {
    addToCartCallback: (item) => console.dir(item),
    resultCardRegularPriceKey: "price",
    resultCardSalePriceKey: "salePrice"
  }
});
```

```js
import CioQuiz from "./constructorio-ui-quizzes-bundled";

CioQuiz({
  selector: "#app",
  includeCSS: false,
  apiKey: "",
  quizId: "coffee-quiz",
  resultsPageOptions: {
    addToCartCallback: (item) => console.dir(item),
    resultCardRegularPriceKey: "price",
    resultCardSalePriceKey: "salePrice"
  }
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,15 +13,15 @@ A UI Library that provides React components to manage the fetching and rendering
 
 ![Quizzes-UI-Example](assets/coffee-quiz.gif)
 
-## Install
+## Installation
 
 ```bash
 npm i @constructor-io/constructorio-ui-quizzes
 ```
 
-## How to use this UI Library
+## Usage
 
-### Import as a React Component
+### Using the React Component
 
 The `CioQuiz` component handles state management, data fetching, and rendering logic for the entire quiz.
 
@@ -35,7 +35,7 @@ function YourComponent() {
     </div>
   );
 ```
-### Import Javascript Bundle
+### Using the Javascript Bundle
 
 This is a framework agnostic method that can be used in any JavaScript project. The `CioQuiz` function provides a simple interface to inject an entire Quizzes UI into the provided `selector`.
 In addition to [Quiz component props](https://constructor-io.github.io/constructorio-ui-quizzes/?path=/docs/quiz-component--docs), this function also accepts `selector` and `includeCSS`.
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ function YourComponent() {
 ```
 ### Import Javascript Bundle
 
-This is a framework agnostic method that can be used in any JavaScript project. The `CioQuiz` function will mount a Quiz to the element for the given selector and it will handle state management, data fetching, and rendering logic for the entire quiz.
+This is a framework agnostic method that can be used in any JavaScript project. The CioQuiz function provides a simple interface to inject an entire Quizzes UI into the provided `selector`.
 In addition to [Quiz component props](https://constructor-io.github.io/constructorio-ui-quizzes/?path=/docs/quiz-component--docs), this function also accepts `selector` and `includeCSS`.
 
 ```js
@@ -47,7 +47,7 @@ CioQuiz({
   selector: "#quiz-container",
   quizId: "coffee-quiz",
   apiKey: "key_wJSdZSiesX5hiVLt",
-  includeCSS: true,
+  includeCSS: true, // Include the default CSS styles. Defaults to true.
   resultsPageOptions: {
     addToCartCallback: (item) => console.dir(item),
     resultCardRegularPriceKey: "price",
--- a/README.md
+++ b/README.md
@@ -13,13 +13,15 @@ A UI Library that provides React components to manage the fetching and rendering
 
 ![Quizzes-UI-Example](assets/coffee-quiz.gif)
 
-## How to use this UI Library
-
-### Install
+## Install
 
 ```bash
 npm i @constructor-io/constructorio-ui-quizzes
 ```
+
+## How to use this UI Library
+
+### Import as a React Component
 
 The `CioQuiz` component handles state management, data fetching, and rendering logic for the entire quiz.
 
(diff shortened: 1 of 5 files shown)


jjl014 (commented):
Nice work! This is looking great from what I can tell! 🚀 

Left a couple of comments.

src/bundled.jsx
```diff
@@ -0,0 +1,49 @@
...
+          <CioQuiz
+            quizId={quizId}
+            apiKey={apiKey}
+            cioJsClient={cioJsClient}
+            resultsPageOptions={resultsPageOptions}
+            quizVersionId={quizVersionId}
+          />
+        </React.StrictMode>
+      );
+    }
```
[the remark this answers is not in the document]
esezen: Good idea. I added an error message. Feel free to edit the message directly if you can think of a better one

vite.config.js
```diff
@@ -0,0 +1,20 @@
...
+    styleId: 'cio-quizzes-styles'
+  })],
+  build: {
+    rollupOptions: {
+      preserveEntrySignatures: 'strict',
+      input: {
+        app: "./src/bundled.jsx",
+      },
+      output: {
+        sourcemap: false,
```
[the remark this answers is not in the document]
esezen: Good catch

src/bundled.jsx
```diff
@@ -0,0 +1,49 @@
...
+  apiKey,
+  quizId,
+  cioJsClient,
+  resultsPageOptions,
+  quizVersionId,
+  includeCSS = true,
+}) => {
+  if (document) {
+    if (!includeCSS) {
+      try {
```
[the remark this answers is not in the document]
esezen: I added it last minute just being paranoid. I wanted to make sure we don't error out in any circumstance but you are right it's unnecessary

src/bundled.jsx
```diff
@@ -0,0 +1,49 @@
...
+  includeCSS = true,
+}) => {
+  if (document) {
+    if (!includeCSS) {
+      const stylesheet = document.getElementById('cio-quizzes-styles');
+
+      if (stylesheet) {
+        stylesheet.disabled = true;
+      }
+    }
```
jjl014, line 22: Forgot to mention, but if we're doing it this way, should we also have an else statement here with:

`document.getElementById('cio-quizzes-styles').disabled = false`

jjl014 (approved):
Thanks for making the changes. This looks great to me! 🚀

esezen:
Enes to update README with bundled approach examples

README.md
```diff
@@ -33,6 +35,26 @@ function YourComponent() {
     </div>
   );
 ```
+### Import Javascript Bundle
+
+This is a framework agnostic method that can be used in any JavaScript project. The `CioQuiz` function will mount a Quiz to the element for the given selector and it will handle state management, data fetching, and rendering logic for the entire quiz.
```
esezen, line 40: I don't love

> The `CioQuiz` function will mount a Quiz to the element for the given selector and it will handle state management, data fetching, and rendering logic for the entire quiz.

but I couldn't find anything better. @jjl014 feel free to edit directly

src/bundled.jsx
```diff
@@ -0,0 +1,38 @@
...
+    ReactDOM.createRoot(containerElement).render(
+      <React.StrictMode>
+        <CioQuizComponent {...rest} />
+      </React.StrictMode>
+    );
+  }
+};
+
+if (window) {
+  window.CioQuiz = CioQuiz;
```
esezen, line 35: Decided to make this `CioQuiz` to make it consistent with the component name

src/bundled.jsx
```diff
@@ -0,0 +1,38 @@
...
+      if (!includeCSS) {
+        stylesheet.disabled = true;
+      } else {
+        stylesheet.disabled = false;
+      }
+    }
+
+    ReactDOM.createRoot(containerElement).render(
+      <React.StrictMode>
+        <CioQuizComponent {...rest} />
```
esezen, line 28: Added `{...rest}` here so we don't forget to add new props to this when we make changes to the component

README.md
jjl014: Maybe we could re-use the description from our docs:

> Allows for a simple interface to inject a quizzes UI into the DOM directly simply by supplying a container reference.

But reword it to fit a bit better for the repo:

> The `CioQuiz` function provides a simple interface to inject an entire Quizzes UI into the provided `selector`.

Probably no necessary to mention state management, data fetching, etc here. Maybe it could be something we can add to the library description at the top of the README instead

README.md
```diff
@@ -33,6 +35,26 @@ function YourComponent() {
...
+In addition to [Quiz component props](https://constructor-io.github.io/constructorio-ui-quizzes/?path=/docs/quiz-component--docs), this function also accepts `selector` and `includeCSS`.
+
+```js
+import CioQuiz from "@constructor-io/constructorio-ui-quizzes/constructorio-ui-quizzes-bundled";
+
+CioQuiz({
+  selector: "#quiz-container",
+  quizId: "coffee-quiz",
+  apiKey: "key_wJSdZSiesX5hiVLt",
+  includeCSS: true,
```
jjl014, line 50: Should we add in a little comment here to explain what this does? For example:

```
includeCSS: true, // Include the default CSS styles. Defaults to true.
```

src/bundled.jsx
```diff
@@ -0,0 +1,38 @@
...
+import CioQuizComponent from './components/CioQuiz';
+import './styles.css';
+
+const CioQuiz = ({ selector, includeCSS = true, ...rest }) => {
+  if (document) {
+    const stylesheet = document.getElementById('cio-quizzes-styles');
+    const containerElement = document.querySelector(selector);
+
+    if (!containerElement) {
+      console.error(`ConstructorioQuizzes: There were no elements found for the provided selector`);
```
jjl014, line 13: Let's update this to `CioQuiz: There were....` since we're changing the window var

src/bundled.jsx
jjl014: Would there be any issues if the consumer passes in random "props"? 🤔

jjl014 (commented):
Nice, thanks for the updates!

Left some comments and questions.

PR #52, opened 2023-04-18
[CSL-2305] Make `resultsPageOptions` required

esezen:


mocca102 (commented):
Can we follow a pattern of having all console errors into a hook called console errors or something like that and then just calling the hook. `useConsoleErrors()`

It just leaves the component cleaner

mocca102:
> Can we follow a pattern of having all console errors into a hook called console errors or something like that and then just calling the hook. `useConsoleErrors()`
> 
> It just leaves the component cleaner

I agree with Jesse https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/59#discussion_r1171339261

mocca102 (approved):

PR #55, opened 2023-04-24
[CSL-2278] Add question specific classes

stanlp1:


src/styles.css
```diff
@@ -104,6 +104,11 @@
   display: flex;
 }
 
+/* Question */
+[class^='cio-question-id'] {
+  display: flex;
```
stanlp1, line 109: Not having display flex breaks our current styles.

src/components/QuizQuestions/index.tsx
```diff
@@ -12,10 +12,10 @@ export default function QuizQuestions(props: { questionResponse: NextQuestionRes
   const questionTypes = getQuestionTypes(nextQuestion?.type);
 
   return (
-    <>
+    <div className={nextQuestion?.id ? `cio-question-id-${nextQuestion?.id}` : ''}>
```
stanlp1, line 15: I added another div here, should we keep it here or move it into the respective question types?

stanlp1: Moved it in each individual just to be safe, not somehow affect styles with this added div.

esezen (changes requested):
This is looking great. It looks like Quizzes team added the key to individual options of select question as well. Can we add that too?
I don't love the naming of the attribute (`key`). I pinged Quizzes team about that

----------------------

Separately, I know the story says class but did we consider making this a data attribute? Are there any pros/cons for class vs data attribute?

stanlp1:
> This is looking great. It looks like Quizzes team added the key to individual options of select question as well. Can we add that too? ~I don't love the naming of the attribute (`key`). I pinged Quizzes team about that~
> 
> Separately, I know the story says class but did we consider making this a data attribute? Are there any pros/cons for class vs data attribute?

Data attribute makes sense. I think there might be slight benefits for selector speed when using classes but I don't think that matters much. Will update.

esezen (approved):
LGTM!

PR #56, opened 2023-04-25
[CSL-1976] Add tracking to components

esezen:
- Quiz Results Load
- Quiz Result Click
- Quiz Conversion

### DIFF
--- a/src/components/ResultCard/ResultCard.tsx
+++ b/src/components/ResultCard/ResultCard.tsx
@@ -69,6 +69,32 @@ export default function ResultCard(props: ResultCardProps) {
     }
   };
 
+  const resultCardContent = () => (
+    <>
+      <div className='cio-result-card-image'>
+        <img src={result.data?.image_url} alt='product' />
+      </div>
+      <div className='cio-result-card-text'>
+        <p className='cio-result-card-title'>{result.value}</p>
+        <div className='cio-result-card-prices'>
+          {salePrice && <span className='cio-result-card-sale-price'>${salePrice}</span>}
+          {regularPrice && (
+            <span className={`cio-result-card-regular-price${salePrice ? '--strike-through' : ''}`}>
+              ${regularPrice}
+            </span>
+          )}
+        </div>
+      </div>
+      <ResultCtaButton item={result} callback={addToCartCallback} />
+    </>
+  );
+
+  const resultCardContentWithLink = () => (
+    <a className='cio-result-card-anchor' href={result.data?.url}>
+      {resultCardContent()}
+    </a>
+  );
+
   return (
     <div
       onClick={clickHandler}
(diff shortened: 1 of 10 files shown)


src/components/ResultCard/ResultCard.tsx
```diff
@@ -1,31 +1,71 @@
...
+  const { cioClient } = useContext(QuizContext);
   const salePrice = salePriceKey && result?.data?.[salePriceKey];
   const regularPrice = regularPriceKey && result?.data?.[regularPriceKey];
 
-  const clickHandler = () => {
-    if (clickItemCallback && typeof clickItemCallback === 'function') {
-      clickItemCallback(result);
+  const clickItemCallback = () => {
+    /* eslint-disable @typescript-eslint/naming-convention */
+    if (resultsResponse && resultsResponse.request && resultsResponse.response) {
```
esezen, line 31: This code partially repeats. I thought about moving the shared parts to a separate function but I think there is enough difference that makes it enough to keep this as is. Let me know what you guys think

src/components/ResultCard/ResultCard.tsx
```diff
@@ -1,31 +1,71 @@
...
+  const {
+    result,
+    addToCartCallback,
+    clickItemCallback: customClickItemCallback,
+    salePriceKey,
+    regularPriceKey,
+    resultPosition,
+  } = props;
+  const { resultsResponse } = useContext(QuizContext);
+  const { cioClient } = useContext(QuizContext);
```
stanlp1, line 25: 🤔 Is there any reason why we have these on separate lines?

src/components/ResultContainer/ResultContainer.tsx
```diff
@@ -20,6 +20,7 @@ export default function ResultContainer(props: IResultContainerProps) {
     resultCardRegularPriceKey,
   } = options;
   const { resultsResponse } = useContext(QuizContext);
+  const { cioClient } = useContext(QuizContext);
```
stanlp1, line 23: Same as here, why do we have these on separate lines? Am I missing something 🤔

src/components/ResultCard/ResultCard.tsx
```diff
@@ -49,24 +89,20 @@ export default function ResultCard(props: ResultCardProps) {
     </>
```
stanlp1, line 89: Should we pass price to the ResultCtaButton?

stanlp1 (changes requested):
Minor comments, implementation looks good to me.

src/components/ResultCard/ResultCard.tsx
esezen: Not really 🤔 Similar code was on separate lines on existing code so I didn't really think why and followed that. I think they can be on the same line. Will update

src/components/ResultCard/ResultCard.tsx
esezen: Good catch. I had it then I accidentally removed it when I was reorganizing code here [6e20c11](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/56/commits/6e20c117e4144c3d33185457edb5ff5b4d5015c8)

stanlp1 (approved):
Nice lgtm.

PR #57, opened 2023-04-27
[CSL-2334] Check if it’s possible to format resultsoptions better in integration tests

esezen:
Copied from https://github.com/Constructor-io/constructorio-ui-autocomplete/blob/main/src/stories/tests/ComponentTests.stories.tsx


stanlp1 (approved):
Screenshot 2023-04-27 at 10 13 34 AM
This looks good to me. But unrelated, I'm wondering if we should have the whole ciojsclient object here?

Could be kind of confusing, wonder if we could only have the configurable options here.

src/stories/Quiz/Component/index.stories.tsx
```diff
@@ -48,8 +48,19 @@ addComponentStoryDescription(
...
+// @ts-ignore
+delete cioJsClient.search;
+// @ts-ignore
+delete cioJsClient.browse;
+// @ts-ignore
+delete cioJsClient.recommendations;
+
 export const ProvideCIOClientInstance = ComponentTemplate.bind({});
-ProvideCIOClientInstance.args = { quizId, cioJsClient, resultsPageOptions };
+ProvideCIOClientInstance.args = { quizId, resultsPageOptions, cioJsClient };
```
jjl014, line 63: I'm not sure if moving the args here works 🤔 

Might need to make adjustments to the [stringifyWithDefaults](https://github.com/Constructor-io/constructorio-ui-quizzes/blob/d37e3f90ab1c4a13e2faebe0d287daf1de3607a9/src/utils.tsx#L64) util function.

jjl014 (approved):
Looks good to me!

Just left one comment around moving the cioJsClient to the end of options list

PR #59, opened 2023-04-30
[Csl-2332] Add use quiz interface

mocca102:


src/hooks/useQuiz.ts
```diff
@@ -0,0 +1,88 @@
...
+  quizApiState?: QuizAPIReducerState;
+  isFirstQuestion?: boolean;
+  events: UseQuizReturnEvents;
+}
+
+type UseQuiz = (quizProps: IQuizProps) => UseQuizReturn;
+
+const useQuiz: UseQuiz = ({ quizId, apiKey, cioJsClient, quizVersionId, resultsPageOptions }) => {
+  // Log console errors for required parameters quizId and resultsPageOptions
+  useConsoleErrors(quizId, resultsPageOptions);
```
esezen, line 45: Why does this need to be a hook? Can't we just use a regular function?

src/hooks/useQuizTrackingAndCbEvents.ts
```diff
@@ -0,0 +1,183 @@
...
+  );
+
+  const quizBackHandler = useCallback(() => {
+    if (dispatchLocalState) {
+      dispatchLocalState({ type: QuestionTypes.Back });
+    }
+  }, [dispatchLocalState]);
+
+  // Quiz result add to cart tracking event
+  const addToCartClickHandler = useCallback(
```
esezen, line 69: Why is this `useCallback` instead of a normal function? 🤔

src/hooks/useQuizTrackingAndCbEvents.ts
```diff
@@ -0,0 +1,183 @@
...
+      if (isFunction(onAddToCartClick)) {
+        e.stopPropagation();
+        onAddToCartClick(result);
+      }
+    },
+    [quizApiState, cioClient, onAddToCartClick]
+  );
+
+  // Quiz result click tracking event
+  const resultClickHandler = useCallback(
```
esezen, line 111: Why is this `useCallback` instead of a normal function? 🤔

src/hooks/useQuizTrackingAndCbEvents.ts
```diff
@@ -0,0 +1,183 @@
...
+      // User custom callback function
+      if (onQuizResultClick && isFunction(onQuizResultClick)) {
+        onQuizResultClick(result, position);
+      }
+    },
+    [quizApiState, cioClient, onQuizResultClick]
+  );
+
+  // Quiz results loaded
+  useEffect(() => {
```
esezen, line 151: Why is this useCallback instead of a normal function? 🤔

src/hooks/useQuizTrackingAndCbEvents.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  return {
+    addToCartClickHandler,
+    resultClickHandler,
+    quizNextHandler,
+    quizBackHandler,
+  };
+};
+
+export default useQuizTrackingAndCbEvents;
```
esezen, line 183: This file name is already long so would it make sense to make it `useQuizTrackingAndCallbackEvents`?

esezen (commented):
This is looking great!

src/hooks/useQuiz.ts
mocca102: I thought this at first as well but following react best practices by having it within a useEffect that only run when any of it's dependency arrays get updated. If we use a regular function it will be called on every render. We can have the logic in a util function like logConsoleErrors() but we still will call it into a useEffect. If you prefer this better I'm open to change it I don't have a preference just that it has to be called in a useEffect

src/hooks/useQuizTrackingAndCbEvents.ts
mocca102: To avoid re initiating the function on each render or each time the hook gets called. This way the function only get reinitiated if any of it's dependency array items gets updated. like `quizApiState` for example. Happy to go over this in detail in our call

src/components/ResultCard/ResultCard.tsx
```diff
@@ -5,66 +5,31 @@ import QuizContext from '../CioQuiz/context';
...
-        result_position_on_page: resultPosition,
-        num_results_per_page,
-      });
-    }
-
-    if (customClickItemCallback && typeof customClickItemCallback === 'function') {
-      customClickItemCallback(result);
+  const clickHandler = (item: Partial<BrowseResultData>) => {
+    if (resultClickHandler) {
+      resultClickHandler(item, resultPosition);
```
deomsj, line 21: `resultPosition` is already in scope, so we are grabbing it from upper scope rather than passing it to `clickHandler` as an argument. Shall we follow the same pattern for `item` aka `result`?

src/components/SelectTypeQuestion/SelectTypeQuestion.tsx
```diff
@@ -12,16 +12,16 @@ interface Selected {
...
+  const { quizApiState, quizLocalState, quizNextHandler, quizBackHandler, isFirstQuestion } =
     useContext(QuizContext);
   let question;
   let type: `${QuestionTypes}`;
   let hasImages = false;
 
-  if (questionResponse) {
-    question = questionResponse.next_question;
+  if (quizApiState?.quizCurrentQuestion) {
+    question = quizApiState?.quizCurrentQuestion.next_question;
```
deomsj, line 22: `question = quizApiState?.quizCurrentQuestion.next_question;`
=>
`const nextQuestion = quizApiState.quizCurrentQuestion?.next_question;`

src/components/SelectTypeQuestion/SelectTypeQuestion.tsx
```diff
@@ -12,16 +12,16 @@ interface Selected {
...
     type = question.type;
-    hasImages = questionResponse.next_question.options.some(
+    hasImages = quizApiState?.quizCurrentQuestion.next_question.options.some(
```
deomsj, line 24: nextQuestion?.options?.some

src/components/SelectTypeQuestion/SelectTypeQuestion.tsx
```diff
@@ -30,8 +30,9 @@ function SelectTypeQuestion() {
   const isDisabled = Object.keys(selected).length === 0;
 
   useEffect(() => {
-    if (questionResponse?.next_question?.type) {
-      const answers = state?.answerInputs?.[questionResponse.next_question.id] || [];
+    if (quizApiState?.quizCurrentQuestion?.next_question?.type) {
+      const answers =
+        quizLocalState?.answerInputs?.[quizApiState?.quizCurrentQuestion.next_question.id] || [];
```
deomsj, line 35: `quizApiState?.quizCurrentQuestion.next_question.id`
=>
`const nextQuestionID = quizApiState?.quizCurrentQuestion.next_question.id;`

src/components/SelectTypeQuestion/SelectTypeQuestion.tsx
```diff
@@ -63,20 +64,8 @@ function SelectTypeQuestion() {
...
-      quizNextHandler({
-        type: questionType,
-        payload: {
-          questionId: questionResponse?.next_question.id,
-          input: Object.keys(selected).filter((key) => selected[Number(key)]),
-          isLastQuestion: questionResponse.is_last_question,
-        },
-      });
+    if (quizNextHandler && !isDisabled && quizApiState?.quizCurrentQuestion) {
+      quizNextHandler(Object.keys(selected).filter((key) => selected[Number(key)]));
```
deomsj, line 68: `quizNextHandler(Object.keys(selected).filter((key) => selected[Number(key)]));`
=>
```
const ??? = Object.keys(selected).filter((key) => selected[Number(key)]);
quizNextHandler(???);
```

src/hooks/useCioClient.ts
```diff
@@ -4,14 +4,18 @@ import { useMemo } from 'react';
 import { getCioClient } from '../utils';
 
 export type CioClientConfig = { apiKey?: string; cioJsClient?: ConstructorIOClient };
-type UseCioClient = (cioClientConfig: CioClientConfig) => ConstructorIOClient | undefined;
+type UseCioClient = (cioClientConfig: CioClientConfig) => ConstructorIOClient | never;
```
deomsj, line 7: how is dis:
`ConstructorIOClient | never`

different than this:
`ConstructorIOClient`

src/hooks/useQuizTrackingAndCbEvents.ts
```diff
@@ -0,0 +1,183 @@
...
+/* eslint-disable @typescript-eslint/naming-convention */
+import ConstructorIOClient from '@constructor-io/constructorio-client-javascript';
+import { GetBrowseResultsResponseData } from '@constructor-io/constructorio-client-javascript/lib/types';
+import { useCallback, useEffect } from 'react';
+import { ActionAnswerQuestion, QuestionTypes } from '../components/CioQuiz/actions';
+import { QuizAPIReducerState } from '../components/CioQuiz/quizApiReducer';
+import { ResultsPageOptions } from '../components/Results/Results';
+import { isFunction } from '../utils';
+
+const useQuizTrackingAndCbEvents = (
```
deomsj, line 11: there is a lot going on in here. would love to explore options for breaking down this complexity for maintainability purposes

src/hooks/useCioClient.ts
mocca102: ConstructorIOClient | never means that the function might never get to it's last line and even return undefined. Basically I'm communicating the `throw new Error() `line through typescript by specifying type never meaning it might throw and error and never return anything even undefined @deomsj

deomsj (changes requested):
Yes!!!! This is shaping up really nicely.

Functionality is all working great on my end when I test with `npm run dev`

I left a bunch of minor comments throughout this PR. Holla with any Qs and lemme know if you'd like to pair or would like another review at any point.

Awesome work here @mocca102!!!

src/components/SelectTypeQuestion/SelectTypeQuestion.tsx
```diff
@@ -12,35 +12,33 @@ interface Selected {
...
-  const { questionResponse, state, quizNextHandler, quizBackHandler, isFirstQuestion } =
-    useContext(QuizContext);
+  const { state, getNextQuestion, getPreviousQuestion } = useContext(QuizContext);
   let question;
   let type: `${QuestionTypes}`;
   let hasImages = false;
 
-  if (questionResponse) {
-    question = questionResponse.next_question;
+  if (state?.quiz.currentQuestion) {
```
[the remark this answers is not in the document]
mocca102: I left it on purpose so I know that it can't be undefined. because what gets returned from useQuiz in state is populated by me and I want it to always have a `state.quiz`. 

Usually I avoid adding optional chaining while I'm using typescript if not necessary so if a certain value I don't expect it to be undefined typescript would warn me about it. What do you think?

src/hooks/useConsoleErrors.ts
```diff
@@ -0,0 +1,23 @@
...
+      // eslint-disable-next-line no-console
+      console.error('quizId is a required field of type string');
+    }
+
+    if (!resultsPageOptions || Object.keys(resultsPageOptions).length === 0) {
+      // eslint-disable-next-line no-console
+      console.error('resultsPageOptions is a required field of type object');
+    }
+
+    if (resultsPageOptions && !resultsPageOptions?.onAddToCartClick) {
```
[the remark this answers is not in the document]
mocca102: This is checking if we have `resultsPageOptions` is true to also check if it has `onAddToCartClick `

The above is checking if no `resultsPageOptions` just console error. If it's undefined no need to also check `onAddToCartClick ` since it will also be undefined

src/utils.tsx
```diff
@@ -130,3 +106,33 @@ export function getPreferredColorScheme() {
   }
   return colorScheme;
 }
+
+export function isFunction(fn) {
+  return fn && typeof fn === 'function';
+}
+
+const isValueExpression = (exp: FilterExpression): exp is FilterExpressionValue =>
+  'name' in exp && 'value' in exp;
```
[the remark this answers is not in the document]
mocca102: I would love to get the credit but this was done by @esezen I just copied the code to utils. It's called [Type Predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) Just to use it as a type guard 

When we have a parameter like `FilterExpression` that can be one of 4 types. And within our code flow we need to know which one of the 4 types it is before starting accessing properties. We use a type guard just to make sure we're using the correct type 
```
if (isValueExpression(exp)) {
 // By using the type predicate util function now we're sure that `exp` parameter will have a `name` and `value`
 // So we make sure in this block which type of the 4 types we're using
}
```

@deomsj

src/components/SelectTypeQuestion/SelectTypeQuestion.tsx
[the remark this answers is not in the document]
deomsj: works for me. so in this case shall we switch from:
`state?.quiz.currentQuestion` to `state.quiz.currentQuestion`?

deomsj (approved):
nice work on this @mocca102!!!

Code is looking good to me and functionality is passing my manual tests when running storybook locally

PR #65, opened 2023-05-11
fix a11y warnings

mocca102:
- Updated the CTA button background color since the one used wasn't accessible friendly
- Same for the input placeholder text 
- Updated ResultCard to not nest a Button inside a link since turns out this isn't accessible friendly
- CTA btn focus styles

Screenshot 2023-05-11 at 8 05 28 PM


esezen:
> Updated the CTA button background color since the one used wasn't accessible friendly
> Same for the input placeholder text

Does this apply only to disabled style?
What was the exact issue with the color? Contrast? 
We might want to talk to Mariano to see if we can keep the same color and make slight modifications

src/styles.css
```diff
@@ -156,14 +156,15 @@
   background: var(--blue-denim-500);
   border: 2px solid var(--blue-denim-600);
 }
-.cio-question-cta-button:active {
+.cio-question-cta-button:focus {
```
esezen, line 159: I'm not entirely sure if it's this change but it makes the button stay "pressed" after I click and move out
image

src/components/ResultCard/ResultCard.tsx
```diff
@@ -51,18 +51,35 @@ export default function ResultCard(props: ResultCardProps) {
...
   );
 
   return (
-    <div
-      onClick={() => clickHandler()}
-      onKeyDown={(e) => keyDownHandler(e)}
-      className='cio-result-card'
-      role='button'
-      tabIndex={0}>
+    <div className='cio-result-card'>
```
esezen, line 82: Why did we remove the events from here? 

--------------

I think the previous behavior was: 
If there is a custom click handler -> Use that instead of redirecting to URL
If there is no custom click handler -> Redirect to URL

--------------

Current behavior:
If there is a custom click handler -> No click action (because we use `resultCardContent` and div with no listeners)
If there is no custom click handler -> Redirect to URL

--------------

If we update the next line to `customClickItemCallback` instead of `! customClickItemCallback`:
If there is a custom click handler -> Call the custom click handler + redirect to URL (we probably don't want to redirect here)
If there is no custom click handler -> No click action (we want the redirect here)

esezen (changes requested):
Thanks for working on the accessibility. Left some comments. 
Quick question, where did we get these check from? Storybook's accessibility tab?
Are we still planning to use `axe/react` or did we decide not to?

src/styles.css
```diff
@@ -399,13 +410,18 @@
...
 }
 .cio-result-card-title {
   color: var(--gray-dust-500);
   margin: 0;
   margin-bottom: 1rem;
   font-weight: 700;
 }
+a .cio-result-card-title:hover {
+  color: var(--blue-denim-500);
+  text-decoration: underline;
```
esezen, line 423: Is this a requirement? Is there anything else we can do to satisfy this? I don't love this
quiz-a11y

src/styles.css
```diff
@@ -414,9 +430,17 @@
   color: #bf1728;
   margin-right: 0.75rem;
 }
+a .cio-result-card-sale-price:hover {
+  color: var(--blue-denim-500);
+  text-decoration: underline;
```
esezen, line 435: Same as above

src/styles.css
```diff
@@ -414,9 +430,17 @@
...
+}
 .cio-result-card-regular-price {
   color: var(--gray-dust-900);
 }
+a .cio-result-card-regular-price:hover {
+  color: var(--blue-denim-500);
+  text-decoration: underline;
```
esezen, line 442: Same as above

esezen (changes requested):

src/components/ResultCard/ResultCard.tsx
```diff
@@ -46,24 +46,33 @@ export default function ResultCard(props: ResultCardProps) {
...
+      role='button'
+      tabIndex={0}>
+      {resultCardContent()}
+    </div>
+  );
+
   const resultCardContentWithLink = () => (
-    <a className='cio-result-card-anchor' href={result.data?.url}>
+    <a
+      className='cio-result-card-content'
```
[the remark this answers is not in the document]
mocca102: removing `cio-result-card-content` and adding the styles to `cio-result-card-anchor`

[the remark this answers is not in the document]
esezen: Alignment issue is still there in the "Basic Usage" example (the layout with a custom click handler)
image

mocca102:
@esezen Added `.cio-result-card-container` with the same styles instead of using the same `.cio-result-card-anchor`  class for library consumers to have different selectors

src/stories/Quiz/tests/index.stories.tsx
```diff
@@ -142,6 +142,6 @@ e2eInteractionTest.play = async ({ canvasElement }) => {
   expect(document.querySelectorAll('.cio-results-filter-option')?.length).toBeGreaterThan(0);
 
   // Reset button test
-  await userEvent.click(await canvas.findByText('Redo Quiz'));
-  expect(await canvas.findByText('Oh, hi there!')).toBeInTheDocument();
+  // await userEvent.click(await canvas.findByText('Redo Quiz'));
```
esezen, line 145: Why did we disable this?

esezen (approved):
Left one comment. Rest LGTM!

PR #68, opened 2023-05-12
[CSL-2354] Update Storybook version

esezen:
Interaction tests are passing ✅ 
CI tests are passing ✅ 
Docs are building properly ✅


mocca102:
It's not a requirement but just heads ups that the way that stories are exported is different now in the docs with the latest release. The update is backward compatible so it didn't break anything. But it might be worth updating the way we export the stories later if we find a need to.

PR #81, opened 2023-06-29
[CSL-2439] Adds option to pass instructions to control bar

Mudaafi:
Adds a descriptor in the control bar for `Select`-type questions:
- Single Select: "Select one option"
- Multi Select: "Select one or more options"

Figured this might be a more common use-case so I made the solution easily extensible (but not more complicated) in case we'd want to add the ability for Question Components to receive such a prop.

### DIFF
--- a/src/components/ControlBar/ControlBar.tsx
+++ b/src/components/ControlBar/ControlBar.tsx
@@ -24,7 +24,7 @@ function ControlBar(props: ControlBarProps) {
   return (
     <div className='cio-question-buttons-container'>
       {showBackButton && <BackButton onClick={backButtonHandler} />}
-      {instructions && instructions !== '' ? (
+      {instructions ? (
         <div className='cio-question-buttons-container-text'>{instructions}</div>
       ) : (
         ''
--- a/src/components/ControlBar/ControlBar.tsx
+++ b/src/components/ControlBar/ControlBar.tsx
@@ -24,11 +24,7 @@ function ControlBar(props: ControlBarProps) {
   return (
     <div className='cio-question-buttons-container'>
       {showBackButton && <BackButton onClick={backButtonHandler} />}
-      {instructions ? (
-        <div className='cio-question-buttons-container-text'>{instructions}</div>
-      ) : (
-        ''
-      )}
+      {instructions && <div className='cio-question-buttons-container-text'>{instructions}</div>}
       <CTAButton
         disabled={isNextButtonDisabled}
         ctaText={ctaButtonText}
--- a/src/components/ControlBar/ControlBar.tsx
+++ b/src/components/ControlBar/ControlBar.tsx
@@ -8,6 +8,7 @@ interface ControlBarProps {
   backButtonHandler?: () => void;
   nextButtonHandler: () => void;
   ctaButtonText?: string;
+  instructions?: string;
 }
 
 function ControlBar(props: ControlBarProps) {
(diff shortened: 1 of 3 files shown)


src/styles.css
```diff
@@ -192,6 +192,9 @@
   background-color: rgba(255, 255, 255, 0.87);
   align-items: center;
 }
+.cio-question-buttons-container-text {
+  display: none;
+}
```
Mudaafi, line 197: Hide if mobile

src/components/ControlBar/ControlBar.tsx
```diff
@@ -17,11 +18,17 @@ function ControlBar(props: ControlBarProps) {
     ctaButtonText,
     nextButtonHandler,
     isNextButtonDisabled,
+    instructions,
   } = props;
 
   return (
     <div className='cio-question-buttons-container'>
       {showBackButton && <BackButton onClick={backButtonHandler} />}
+      {instructions && instructions !== '' ? (
```
mocca102, line 27: isn't just `instructions` enough? since an empty string is false

mocca102: I think we can just do `{instructions && <div className='cio-question-buttons-container-text'>{instructions}</div>}`

mocca102 (approved):
LGTM! left a small comment

src/components/ControlBar/ControlBar.tsx
```diff
@@ -17,11 +18,17 @@ function ControlBar(props: ControlBarProps) {
     ctaButtonText,
     nextButtonHandler,
     isNextButtonDisabled,
+    instructions,
   } = props;
 
   return (
     <div className='cio-question-buttons-container'>
       {showBackButton && <BackButton onClick={backButtonHandler} />}
+      {instructions ? (
```
esezen, line 27: We should be able to clip this even further like Islam mentioned 
```
{instructions && <div className='cio-question-buttons-container-text'>{instructions}</div>}
```
Not a big deal, it just has a bit better readability

src/components/SelectTypeQuestion/SelectTypeQuestion.tsx
```diff
@@ -16,11 +16,22 @@ function SelectTypeQuestion() {
...
   if (state?.quiz.currentQuestion) {
     question = state.quiz.currentQuestion.next_question;
     type = question.type;
     hasImages = question.options.some((option: QuestionOption) => option.images);
+    switch (type) {
+      case QuestionTypes.MultipleSelect:
+        instructions = 'Select one or more options';
+        break;
+      case QuestionTypes.SingleSelect:
+        instructions = 'Select one option';
```
esezen, line 30: Designs don't have `Select one option` if I'm looking at the correct thing. Do you mind checking in with Mariano to see if that's intentional? If it is, we can remove it from single select

esezen (commented):
This is looking good. Left 2 small comments
Let me know what you think and we can merge 🚀

src/components/SelectTypeQuestion/SelectTypeQuestion.tsx
Mudaafi: Sure thing, I'll drop him a ping

src/components/ControlBar/ControlBar.tsx
Mudaafi: I need to burn this pattern into my brain. Thanks for the callout!

esezen: Not a must have, just makes it more concise

src/components/SelectTypeQuestion/SelectTypeQuestion.tsx
Mudaafi: According to Mariano, it was intentional as the original ux intent for single-select was to nav to the next question immediately when the user selects one option. I've removed the single select and made the code more concise.

PR #82, opened 2023-06-30
[CSL-2303] Fixes layout shifts in Open Text and Cover Questions

Mudaafi:
Note changes to:
- Cover-Questions page (mobile) - Fixed height makes the image larger
- Cover-Question HTML Structure
    - I moved the elements around and added the flex-direction: row-reverse on >tablet-sizes since (a) it felt weird, from a mobile-first standpoint, that the image is lower on the html than the content and (b) allows us to apply alignment correctly on flex-column (i.e. on mobile, content aligns to the top by default rather than to the bottom w/ flex-direction: column-reverse)

### DIFF
--- a/src/components/CoverTypeQuestion/CoverTypeQuestion.tsx
+++ b/src/components/CoverTypeQuestion/CoverTypeQuestion.tsx
@@ -29,6 +29,7 @@ export default function CoverTypeQuestion() {
         cio-cover-question-container${hasImage ? '--with-image' : ''}
       `}
         data-question-key={question.key}>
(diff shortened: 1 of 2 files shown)


mocca102 (commented):
In open text questions the image was on the left now it's on the right for larger screens. Not that I don't like it actually I like the image on the right more. Just raising it

Mudaafi:
Nice catch! I used a Question-specific class to reset the default style to what we expect.

esezen (approved):
LGTM!

PR #87, opened 2023-07-13
Add rating component

mocca102:
This isn't necessarily the final look we can use Mariano's design. This is just adding the parameters


src/components/ResultCard/ResultCard.tsx
```diff
@@ -37,6 +42,15 @@ export default function ResultCard(props: ResultCardProps) {
       </div>
       <div className='cio-result-card-text'>
         <p className='cio-result-card-title'>{result.value}</p>
+        <div className='cio-result-card-rating'>
```
esezen, line 45: This is not a must have, a question.
What do you think about putting rating and prices into a container div? That way if anyone wants to flip their position (price first), they can do so easily with CSS.

esezen (commented):
This is looking good. Left some small comments, let me know what you think. 

Let's get this out and also contact Mariano for the final design so we can update the styles later

src/components/ResultCard/ResultCard.tsx
mocca102: Actually good idea!

src/components/ResultCard/ResultCard.tsx
```diff
@@ -37,6 +42,15 @@ export default function ResultCard(props: ResultCardProps) {
...
+          {ratingScore && (
+            <span className='cio-result-card-rating-score'>
+              {Array(Number(ratingScore)).fill('★')}
+              {Array(5 - Number(ratingScore)).fill('☆')}{' '}
```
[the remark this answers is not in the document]
mocca102: I think it got added by the formatter. Thanks for catching it

esezen (approved):
LGTM!

esezen (approved):
LGTM!

PR #89, opened 2023-07-17
Troubleshooting Javascript syntax version

mocca102:


README.md
```diff
@@ -82,6 +82,18 @@ import '@constructor-io/constructorio-ui-quizzes/styles.css';
 
 ### Known Issues
 
+**Javascript version**
+
+The default exported javascript syntax is ESNext, the latest supported syntax of JavaScript.
```
esezen, line 87: What do you think about this?

This library exports two different versions
* CommonJS (cjs)
* ECMAScript Modules (mjs)

If you are using an `import` statement, `mjs` version will be imported.
If you are using a `require` statement, `cjs` version will be imported.

Since `mjs` version is compiled to `ESNext`, there could be features that are not supported by older Chrome versions (ex. optional chaining). If you are receiving an error similar to the following error, please import the `cjs` version which is compiled to ES2015.

`import CioQuiz from '@constructor-io/constructorio-ui-quizzes/commonJs'`

esezen (commented):
@mocca102  This is looking good. I left some comments, let me know what do you think about them. We can adjust or ignore if you don't like them

package.json
```diff
@@ -26,6 +26,10 @@
     ".": {
       "import": "./lib/mjs/index.js",
       "require": "./lib/cjs/index.js"
+    },
+    "./commonJs": {
```
esezen, line 30: I could swear I left a comment here but it somehow disappeared. 

Do you have a preference `cjs` vs `commonJs`? If not, can we do `cjs`?

esezen (commented):
@mocca102 thanks for making the changes. I still think it's important to mention the differences between `import` and `require` but I might be wrong. Do you mind bringing this to standup tomorrow so we can discuss with the team?

README.md
esezen: > If you are using an import statement, mjs version will be imported.
If you are using a require statement, cjs version will be imported.

Settled on Slack

esezen (approved):
LGTM!

PR #76, opened 2023-06-20
Add storybook hook example

mocca102:

PR #91, opened 2023-07-19
Fix try again button on zero results page

mocca102:
This is fixed in the [hooks PR](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/76/files#diff-3b8a6bcca4206de32dc0813ce044e6086cb3748e9242d5a741b5058a30c65774R19) as well but did a separate fix here also just in case the hooks PR takes longer


esezen:
Can we close this without merging if we merge https://github.com/Constructor-io/constructorio-ui-quizzes/pull/76 today?

esezen:
Fixed in another PR

PR #92, opened 2023-07-19
[CSL-2279] Single select question - Select behavior

esezen:
Changes the select behavior to go to the next question as soon as user clicks on an option


mocca102 (approved):
LGTM!

mocca102 (changes requested):
Can we update the tests to accommodate for the new behavior. Since in the tests for single select questions we try to click the next button after option selection and it's failing because of that

mocca102 (approved):
LGTM!

PR #93, opened 2023-07-19
[CSL-2621] Update multi select instructions

esezen:


src/styles.css
```diff
@@ -360,6 +357,16 @@
...
 }
 
+/* Select Instructions */
+.cio-select-question-instructions {
+  text-align: center;
+  font-size: 1.125rem;
+  color: var(--gray-dust-100);
+  line-height: 150%;
+  margin-top: 2.75rem;
+  margin-bottom: -1rem;
```
esezen, line 367: This is to adjust the space between description and options in both single select and multi select. Let me know if negative margins would cause an issue

src/styles.css
```diff
@@ -360,6 +357,16 @@
   margin-bottom: 0;
 }
 
+/* Select Instructions */
+.cio-select-question-instructions {
+  text-align: center;
+  font-size: 1.125rem;
+  color: var(--gray-dust-100);
+  line-height: 150%;
+  margin-top: 2.75rem;
```
mocca102, line 366: I couldn't find the space pixels exactly from the designs but just by looking the margin-top seems a bit too much

mocca102 (approved):
LGTM! Left a small comment

src/components/SelectTypeQuestion/SelectTypeQuestion.tsx
```diff
@@ -112,7 +112,6 @@ function SelectTypeQuestion() {
           backButtonHandler={previousQuestion}
           showBackButton={!state?.quiz.isFirstQuestion}
           ctaButtonText={question?.cta_text}
-          instructions={instructions}
```
mocca102, line 114: This isn't used anymore so just removed it.

mocca102: Also created this story to create github action to check types to catch things like this later https://linear.app/constructor/issue/CSL-2626/os-ui-library-add-github-action-to-check-types

PR #98, opened 2023-07-24
[CSL-2600] Remove top margin on small tablet view

esezen:
Before
image
After
image

### DIFF
--- a/src/styles.css
+++ b/src/styles.css
@@ -561,7 +561,6 @@
   /* Container */
   .cio-container--with-image {
     display: flex;
-    justify-content: center;
     align-items: center;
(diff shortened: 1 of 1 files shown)


src/styles.css
```diff
@@ -578,7 +577,6 @@
 
   /* Cover Page Component */
   .cio-cover-question-container--with-image {
-    justify-content: center;
```
esezen, line 581: Since the same style comes from `cio-container--with-image`, I believe it's redundant here. Let me know if there is an edge case this handles

mocca102 (approved):
LGTM!

PR #99, opened 2023-07-24
[CSL-2632] - Open results in a new tab

HHHindawy:


src/components/ResultCard/ResultCard.tsx
```diff
@@ -62,8 +62,10 @@ export default function ResultCard(props: ResultCardProps) {
 
   const resultCardContentWithLink = () => (
     <a
+      rel='noreferrer'
```
mocca102, line 65: Can we discuss with the team if we want noreferrer here?

HHHindawy: Sure, the thing is we shouldn't specify a `_blank` target attribute without setting the `rel` to either `noopener` or `noreferrer` (Security risk).

esezen: We definitely need one of those. `noopener` might be the better choice here

HHHindawy: I don't think `noopener` is enough:

image

@esezen @mocca102

esezen: Interesting 🤔 Ok let's keep it as is. @HHHindawy can you resolve the conflicts? I can merge this today

HHHindawy: @esezen Done 👍

esezen (approved):
LGTM!

PR #103, opened 2023-07-25
[NO-CSL] Update hooks story

esezen:


src/stories/Quiz/Hooks/index.tsx
```diff
@@ -105,7 +103,7 @@ export default function HooksTemplate(args) {
     }
 
     const { currentQuestion } = state.quiz;
-    const currentQuestionData = state.quiz.currentQuestion.next_question;
+    const currentQuestionData = state?.quiz?.currentQuestion?.next_question;
```
deomsj, line 106: this can be simplified:
```
    const { currentQuestion } = state.quiz;
    const currentQuestionData = state?.quiz?.currentQuestion?.next_question;
```

also... 
`state?.quiz?.` implies that it is possible for `state.quiz` to be undefined here....
but `const { currentQuestion } = state.quiz;` will throw an error if state.quiz is undefined

mayhaps we are using `?.` a bit too aggressively, sending mixed signals to devs who may be trying to interpret this code

src/stories/Quiz/Hooks/index.tsx
```diff
@@ -105,7 +103,7 @@ export default function HooksTemplate(args) {
...
     if (currentQuestion) {
```
deomsj, line 107: i'm not sure, but can we check for `currentQuestionData` here instead of `currentQuestion`, then we wouldn't need all of the `?.` in any of the nested JSX. we could even destructure  `currentQuestionData` after we have checked that it is defined in order to make the JSX more readable

deomsj (changes requested):
Screenshot 2023-07-25 at 3 53 27 PM

src/stories/Quiz/Hooks/index.tsx
```diff
@@ -164,10 +162,10 @@ export default function HooksTemplate(args) {
                 <p>{currentQuestionData.description}</p>
               </div>
               <div className='cio-question-options-container'>
-                {currentQuestion.next_question?.options?.map((option: QuestionOption) => (
+                {currentQuestionData.options.map((option: QuestionOption) => (
```
deomsj, line 165: double checking, @esezen : will `currentQuestionData.options` always be an array whenever `currentQuestionData` is truthy?

PR #104, opened 2023-07-25
[CSL-2577] - Configure showing session modal conditionally after reaching results step

HHHindawy:


esezen (approved):
LGTM! Let's get another review from @mocca102

src/stories/Quiz/Component/QuizResults/QuizResults.stories.tsx
```diff
@@ -37,8 +37,8 @@ export const QuizResultsPage: Story = {
     // 👇 Assert DOM structure
     const resultsTitle = canvas.getByText('Here are your results');
     const selectedOptions = document.querySelector('.cio-results-filter-options');
-    const resultsFilterChocolate = canvas.getByText('Chocolate');
-    const resultsFilterMedium = canvas.getByText('Medium');
+    const resultsFilterChocolate = canvas.getByText('Option 1');
+    const resultsFilterMedium = canvas.getByText('Option 2');
```
mocca102, line 41: Same here lets revert it back to "Medium"

src/hooks/useQuizEvents/index.ts
```diff
@@ -23,6 +23,7 @@ const useQuizEvents: UseQuizEvents = (quizOptions, cioClient, quizState) => {
     dispatchLocalState,
     dispatchApiState,
     hasQuizStoredState,
+    isQuizCompleted,
```
mocca102, line 26: If you notice here we're just passing `isQuizCompleted` to `useQuizEvents` and then return it as is. I think you did this following what is done for `hasQuizStoredState` But I think it's not necessary for it as well

Can we return those values from useQuizLocalState directly without passing them to useQuizEvents?

src/components/CioQuiz/index.tsx
```diff
@@ -31,13 +31,19 @@ export default function CioQuiz(props: IQuizProps) {
   const [showSessionPrompt, setShowSessionPrompt] = useState(false);
   const { resultsPageOptions, sessionStateOptions } = props;
 
+  const skipToResults = isQuizCompleted() && !sessionStateOptions?.showSessionModalOnResults;
```
mocca102, line 34: I think all we care about here is whether we will `skipToResults` or not. We don't care about exposing `isQuizCompleted` or `hasSessionStorageState` 

What do you think if we moved this logic inside the `useQuizLocalState` hook and only returned whether we should `skipToResults` or not?

mocca102 (changes requested):

src/components/CioQuiz/index.tsx
HHHindawy: I like that 👍

mocca102 (approved):
LGTM!

PR #106, opened 2023-07-27
[CSL-2242] Add background images to select questions

esezen:


src/components/CioQuiz/index.tsx
```diff
@@ -80,6 +88,9 @@ export default function CioQuiz(props: IQuizProps) {
         <QuizContext.Provider value={contextValue}>
           {state.quiz.results && <ResultContainer options={resultsPageOptions} />}
           {state.quiz.currentQuestion && <QuizQuestions />}
+          {state.quiz.currentQuestion && (
+            <ControlBar ctaButtonText={questionData?.cta_text || undefined} />
```
esezen, line 92: Typescript wasn't happy with `questionData?.cta_text` here, let me know if there is a better way to do this

stanlp1 (approved):
LGTM. Everything appears to work as expected.

src/components/CioQuiz/index.tsx
```diff
@@ -87,7 +95,12 @@ export default function CioQuiz(props: IQuizProps) {
           {state.quiz.results || state.quiz.skipToResults ? (
             <ResultContainer options={resultsPageOptions} />
           ) : (
-            state.quiz.currentQuestion && <QuizQuestions />
+            state.quiz.currentQuestion && (
+              <>
+                <QuizQuestions />
+                <ControlBar ctaButtonText={questionData?.cta_text || undefined} />
```
deomsj, line 101: `questionData?.cta_text || undefined`

could be simplified to: `questionData?.cta_text`

right?

src/styles.css
```diff
@@ -86,7 +92,15 @@
   padding-bottom: var(--bottom-bar-height);
   position: relative;
   align-self: start;
-  min-height: 100%;
+  height: calc(100% - calc(var(--bottom-bar-height))); /* Full height - margin - control bar */
```
deomsj, line 95: css math. nice

src/styles.css
```diff
@@ -302,12 +318,15 @@
...
   flex-direction: column;
   align-items: center;
-  align-self: start;
   max-width: 1200px;
   padding: 1rem;
   margin-top: 2rem;
-  min-height: calc(100% - 3rem); /* Full height - padding - margin */
+  min-height: calc(
+    100% - 2rem - calc(var(--bottom-bar-height))
+  ); /* Full height - padding - margin */
```
deomsj, line 326: can we simplify:
```
calc(
    100% - 2rem - calc(var(--bottom-bar-height))
  )
```
  
to: `calc(100% - 2rem - var(--bottom-bar-height))`

src/styles.css
```diff
@@ -623,6 +644,7 @@
   .cio-container {
     align-items: center;
     margin-top: 0;
+    height: calc(100% - calc(var(--bottom-bar-height))); /* Full height - control bar */
```
deomsj, line 647: same as above, do we need two different calc functions in this expression?

src/styles.css
```diff
@@ -686,9 +698,11 @@
   }
   .cio-select-question-container {
     justify-content: center;
-    margin: 0;
+    margin-bottom: 0;
+    margin-top: 0;
     padding: 2rem;
     padding-bottom: 0;
+    min-height: calc(100% - calc(var(--bottom-bar-height))); /* Full height - padding - margin */
```
deomsj, line 705: same as above

deomsj (commented):
nice. just a few comments on potential opportunities to simplify code

src/components/CioQuiz/index.tsx
esezen: Typescript gets really mad because that returns null and null is not undefined

PR #107, opened 2023-07-28
[CSL-2597] Add to favorites button

stanlp1:


esezen (commented):
This is looking great. Can you add `onAddToFavoritesClick` to `functionStrings` in `utils`?
image

Also do you mind reaching out to Mariano to see if he is happy with the size of the icon? We don't need to wait for his response, we can merge this and change the size in a different PR if needed

src/components/CioQuiz/index.tsx
```diff
@@ -56,6 +58,7 @@ export default function CioQuiz(props: IQuizProps) {
     getResetQuizButtonProps,
     getSelectInputProps,
     customClickItemCallback: !!resultsPageOptions?.onQuizResultClick,
+    customAddToFavoritesCallback: !!resultsPageOptions?.onAddToFavoritesClick,
```
mocca102, line 61: Why add `customAddToFavoritesCallback` to the customer facing interface? I'm not opposed to using it just why return it in the hook? I think we can remove it along with `customClickItemCallback`

mocca102 (changes requested):
Can we add another SVG with a filled heart when add_to_favorites fire and only fire the event if the favorites icon isn't active/clicked already. Basically like a toggle since currently the icon doesn't change on click and it looks as if it's not working

src/components/CioQuiz/index.tsx
mocca102: Ignore this please

src/components/ResultFavoritesButton/ResultFavoritesButton.tsx
```diff
@@ -8,14 +8,22 @@ interface ResultFavoritesButtonProps {
 }
 
 export default function ResultFavoritesButton(props: ResultFavoritesButtonProps) {
+  const [isFavorited, setIsFavorited] = useState(false);
   const { item, price } = props;
   const { getAddToFavoritesButtonProps } = useContext(QuizContext);
 
+  const clickHandler = () => {
```
mocca102, line 15: Lets not have any logic in presentational components and have the logic in hooks

src/components/ResultFavoritesButton/ResultFavoritesButton.tsx
```diff
@@ -8,14 +8,22 @@ interface ResultFavoritesButtonProps {
 }
 
 export default function ResultFavoritesButton(props: ResultFavoritesButtonProps) {
+  const [isFavorited, setIsFavorited] = useState(false);
```
mocca102, line 11: Lets not have any logic in presentational components and have the logic in hooks

mocca102:
Thanks for making the changes, looking way better. Just a couple of things
- Misspelling in file name useAddToFavoritesbuttonProps -> useAddToFavoritesButtonProps
- Conversion event should only fire if favorites button isn't already clicked
- Not sure if it's this PR but single result card is misaligned

mocca102 (changes requested):
Thanks for making the changes, looking way better. Just a couple of things

- Misspelling in file name useAddToFavoritesbuttonProps -> useAddToFavoritesButtonProps
- Conversion event should only fire if favorites button isn't already clicked
- Not sure if it's this PR but single result card is misaligned
Screenshot 2023-08-03 at 4 22 54 PM

mocca102:
Ignore the visual bug. Enes fixed it in another PR @stanlp1

mocca102 (approved):
LGTM!

PR #110, opened 2023-08-03
[NO-CSL] Simplify styles

esezen:


src/styles.css
```diff
@@ -532,11 +528,6 @@
 .cio-result-card-rating-score {
   margin-right: 5px;
 }
-.cio-result-card-discount {
```
esezen, line 535: @mocca102 I moved this to Storybook styles since we don't support it natively

mocca102 (approved):
LGTM!

PR #111, opened 2023-08-04
Fix layout shift on add to cart hover

esezen:


stanlp1 (approved):
Seems like that fixed it.

One thing I noticed is in our result card stories, the add to cart buttons use cio-question-cta-button, not cio-result-card-cta-button.

Is that intentional?

https://github.com/Constructor-io/constructorio-ui-quizzes/blob/0ff850074ccfc81a4c007897b25cd845c139794e/src/stories/Quiz/tests/mocks.ts#L187

esezen:
> Is that intentional?

I don't think so, fixed!

PR #114, opened 2023-08-15
Csl 2680 os UI library quizzes implement visual tests using chromatic

mocca102:
To make this PR small I only did refactor the CoverQuestion to have the new Kitchen Sinks pattern

- Added chromatic
- Added chromatic github action to automatically deploy to chromatic on pull requests
- Component Story will have a primary Component and Variations of that component in every possible container size
- Removed tests from CoverQuestion Story since visual tests will replace it
- All components added starting with Story* are just to preview stories. No new functionality 

Chromatic app link https://www.chromatic.com/builds?appId=64d3cc5a5b332c388f6e5630


src/styles.css
```diff
@@ -167,7 +167,7 @@
 /* Image */
 .cio-question-image {
   width: 100%;
-  height: 100%;
+  max-height: 100%;
```
mocca102, line 170: This was causing an issue with the variation for Big table. Remove it and you'll be able to reproduce the css bug

src/components/CoverTypeQuestion/CoverTypeQuestion.tsx
```diff
@@ -1,35 +1,34 @@
...
 import QuestionTitle from '../QuestionTitle/QuestionTitle';
-import QuizContext from '../CioQuiz/context';
 import QuestionDescription from '../QuestionDescription/QuestionDescription';
 import { renderImages } from '../../utils';
+import { QuestionImages } from '../../types';
 
-export default function CoverTypeQuestion() {
-  const { state } = useContext(QuizContext);
-  let question;
+export interface ICoverTypeQuestionProps {
```
mocca102, line 8: Making it clearer what CoverQuestion props are and passing them as props instead of depending on context in case a user want to use the component without necessarily having to implement context for it to work

.storybook/main.js
```diff
@@ -16,6 +16,6 @@ module.exports = {
   docs: {
     //👇 See the table below for the list of supported options
     autodocs: 'tag',
-    defaultName: 'Documentation',
+    defaultName: 'Docs',
```
esezen, line 19: Just curious, why?

.storybook/storybook-styles.css
```diff
@@ -1,18 +1,25 @@
 .docs-story > div:first-child {
   min-height: 500px;
-  background-color: #212425;
+  background-color: #F8F8F8;
```
esezen, line 3: Again, not important, why?

src/components/QuizQuestions/index.tsx
```diff
@@ -10,11 +10,23 @@ export default function QuizQuestions() {
...
-      {questionTypes.isOpenQuestion && <OpenTextQuestion key={nextQuestion?.id} />}
-      {questionTypes.isCoverQuestion && <CoverTypeQuestion key={nextQuestion?.id} />}
-      {questionTypes.isSelectQuestion && <SelectTypeQuestion key={nextQuestion?.id} />}
-    </>
-  );
+  if (nextQuestion) {
+    return (
+      <>
+        {questionTypes.isOpenQuestion && <OpenTextQuestion key={nextQuestion?.id} />}
+        {questionTypes.isCoverQuestion && (
```
esezen, line 17: Why does CoverQuestion take a bunch of props but others rely on context?

src/components/CoverTypeQuestion/CoverTypeQuestion.tsx
esezen, line 8: How would this work without the context and why are we only doing it for Cover?

src/stories/Quiz/Component/QuestionTypes/CoverQuestion.stories.tsx
```diff
@@ -1,48 +1,49 @@
-import { expect } from '@storybook/jest';
-import { CoverQuestion } from '@constructor-io/constructorio-client-javascript/lib/types';
 import type { Meta, StoryObj } from '@storybook/react';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
esezen, line 2: We have `import/no-extraneous-dependencies` in our eslint, why do we need it here?

src/styles.css
esezen: I tried it but couldn't reproduce, do you mind posting an example?

esezen (commented):
I like the Kitchen Sink concept and it works fine for dashboard because they are using it internally but
1. I think this could be confusing for users of the library. Is there a way to hide these in the actual docs and only use them for Chromatic
2. This only exists for one question, are we planning to do it for all questions if we decide to use it?

Chromatic UI Tests make sense. Do we need Storybook Publish? Or do they come together?

.storybook/main.js
mocca102: less characters 😅

src/components/QuizQuestions/index.tsx
mocca102: Just refactored one component so the PR is not big and also to make sure we agree on those changes before changing all of the components

src/components/CoverTypeQuestion/CoverTypeQuestion.tsx
mocca102: Just refactored one component so the PR is not big and also to make sure we agree on those changes before changing all of the components

.storybook/storybook-styles.css
mocca102: It was fixing something but seems it's not anymore. removed

esezen:
Summary of discussion

1. Let's rename "Kitchen Sink"
2. Let's revert the component props changes for now

esezen (approved):
LGTM!

PR #117, opened 2023-08-22
[CSL-2648] Storybook Docs - Retrieving Open Text Answers

Mudaafi:
- Creates a new folder for "Specific Examples" to demonstrates how to achieve certain, specific outcomes
- Example uses `QuizContext` and `ControlBar` component, both of which aren't currently exposed. They aren't required to access the Open Text Answers but it makes sense to expose them. Will bring it up in a separate PR.
- Wanted to do something like `export RetrievingAnswers from './RetrievingAnswers'` in index.stories.tsx but doing this doesn't space-delimit the Story in the Storybook docs


mocca102 (approved):
LGTM!

PR #122, opened 2023-08-24
[CSL-2715] onQuizNextQuestion callback

esezen:
I decided to leave the callbacks under `resultsPageOptions` as is because they are highly tied to results page. I think we can leave them there and add the rest of the callbacks to `callbacks` object. Let me know if you think otherwise, open to suggestions

### DIFF
--- a/src/hooks/useQuizEvents/useQuizNextClick.ts
+++ b/src/hooks/useQuizEvents/useQuizNextClick.ts
@@ -2,26 +2,39 @@ import { useCallback } from 'react';
 import { ActionAnswerQuestion, QuestionTypes } from '../../components/CioQuiz/actions';
 import { QuizAPIReducerState } from '../../components/CioQuiz/quizApiReducer';
 import { QuizLocalReducerState } from '../../components/CioQuiz/quizLocalReducer';
-import { QuizEventsReturn } from '../../types';
+import { QuizEventsReturn, OnQuizNextQuestion } from '../../types';
+import { isFunction } from '../../utils';
 
 const useQuizNextClick = (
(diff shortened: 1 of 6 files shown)


src/hooks/useQuizEvents/useQuizNextClick.ts
```diff
@@ -2,26 +2,39 @@ import { useCallback } from 'react';
...
       const currentAnswerInput = quizLocalState.answerInputs[currentQuestionId];
+
       if (currentAnswerInput?.value?.length || currentQuestion?.type === 'cover') {
         dispatchLocalState({
           type: QuestionTypes.Next,
           payload: quizApiState.quizCurrentQuestion,
         });
       }
+
+      if (currentQuestion && onQuizNextQuestion && isFunction(onQuizNextQuestion)) {
```
stanlp1, line 28: This is a nitpick. but isFunction already checks onQuizNextQuestion so we don't need to check && onQuizNextQuestion here again 😅

stanlp1 (approved):
I have a nitpick comment but otherwise lgtm.

src/hooks/useQuizEvents/useQuizNextClick.ts
stanlp1: All good if typescript yells. I think we can add ! on the quiz function when you call it like onQuizNextQuestion!(...)

But both are fine.

esezen: You are a wizard @stanlp1 🧙

mocca102 (approved):
LGTM!

PR #126, opened 2023-09-10
Update full quiz story docs

mocca102:
- Updated callbacks key to include all callbacks
- New docs layout
- Update callbacks, resultPageOptions and resultCardOptions


.storybook/main.js
```diff
@@ -27,4 +27,19 @@ module.exports = {
     ...config,
     LOGGER: true,
   }),
+  typescript: {
```
mocca102, line 30: Options to autogenerate types from typescript
It works for direct props of the component and not nested types. Thus the usage of argTypes as well to document the nested types

.storybook/preview.jsx
```diff
@@ -22,13 +22,7 @@ export const parameters = {
...
   },
   docs: {
-    page: () => (
-      <>
-        <Title />
-        <Description />
-        <Stories includePrimary />
-      </>
-    )
+    toc: true,
```
mocca102, line 25: New navigation for page table of contents 
Screenshot 2023-09-10 at 10 10 22 PM

package.json
```diff
@@ -66,14 +66,14 @@
   "devDependencies": {
     "@axe-core/react": "^4.7.0",
     "@cspell/eslint-plugin": "^6.18.1",
-    "@storybook/addon-a11y": "^7.0.11",
-    "@storybook/addon-actions": "7.0.11",
-    "@storybook/addon-essentials": "7.0.11",
-    "@storybook/addon-interactions": "^7.0.11",
-    "@storybook/addon-links": "7.0.11",
-    "@storybook/addon-docs": "7.0.11",
+    "@storybook/addon-a11y": "^7.4.0",
```
mocca102, line 69: Updated storybook

src/types.ts
```diff
@@ -59,18 +56,22 @@ export type QuestionWithAnswer = Question & {
   answer: AnswerInput;
 };
 
-export type OnQuizNextQuestion = (question: QuestionWithAnswer) => void;
 
 export interface Callbacks {
-  onQuizNextQuestion?: OnQuizNextQuestion;
+  onQuizNextQuestion?: QuizResultsEventsProps.OnQuizNextQuestion;
+  onQuizResultsLoaded?: QuizResultsEventsProps.OnQuizResultsLoaded;
```
mocca102, line 62: Move the rest of callbacks from resultsPageOptions to callbacks

src/types.ts
```diff
@@ -59,18 +56,22 @@ export type QuestionWithAnswer = Question & {
...
 }
 
 export interface IQuizProps {
   apiKey?: string;
   cioJsClient?: ConstructorIOClient;
   quizId: string;
   quizVersionId?: string;
-  resultsPageOptions: ResultsPageOptions;
+  resultsPageOptions?: ResultsPageOptions;
+  resultCardOptions?: ResultCardOptions;
```
mocca102, line 74: Added ResultCardOptions to include result card related props

src/hooks/useConsoleErrors.ts
```diff
@@ -2,23 +2,18 @@ import { useEffect } from 'react';
...
-  const { quizId, resultsPageOptions } = quizOptions;
+  const { quizId, resultsPageOptions, callbacks } = quizOptions;
   useEffect(() => {
     if (!quizId) {
       // eslint-disable-next-line no-console
       console.error('quizId is a required field of type string');
     }
 
-    if (!resultsPageOptions || Object.keys(resultsPageOptions).length === 0) {
+    if (!callbacks?.onAddToCartClick) {
```
mocca102, line 12: Removed resultsPageOptions since they're only required for the component interface and not for hooks.
Added to the docs though

src/hooks/useConsoleErrors.ts
```diff
@@ -2,23 +2,18 @@ import { useEffect } from 'react';
...
 const useConsoleErrors = (quizOptions: IQuizProps) => {
-  const { quizId, resultsPageOptions } = quizOptions;
+  const { quizId, resultsPageOptions, callbacks } = quizOptions;
   useEffect(() => {
     if (!quizId) {
       // eslint-disable-next-line no-console
       console.error('quizId is a required field of type string');
     }
 
-    if (!resultsPageOptions || Object.keys(resultsPageOptions).length === 0) {
```
mocca102, line 12: Removed resultsPageOptions since they're only required for the component interface and not for hooks.
Added to the docs though

src/types.ts
```diff
@@ -39,14 +39,11 @@ export namespace QuizResultsEventsProps {
   export type OnQuizResultClick = (result: QuizResultDataPartial, position: number) => void;
   export type OnAddToCartClick = (result: QuizResultDataPartial) => void;
   export type OnAddToFavoritesClick = (result: QuizResultDataPartial) => void;
+  export type OnQuizNextQuestion = (question: QuestionWithAnswer) => void;
 }
 
-export interface ResultsPageOptions extends ResultCardProps {
+export interface ResultsPageOptions {
```
mocca102, line 45: Removed callbacks

.github/workflows/run-license-check.yml
```diff
@@ -20,6 +20,6 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
       - name: Install dependencies
-        run: npm install
+        run: npm ci
```
mocca102, line 23: For consistency

.storybook/preview.jsx
```diff
@@ -21,15 +21,6 @@ export const parameters = {
   previewTabs: {
     'storybook/docs/panel': { index: -1 }
   },
-  docs: {
```
mocca102, line 24: Using auto generated docs with some tweaks instead

.storybook/preview.jsx
mocca102: This is part of the updated storybook version. I removed the update changes since it created a big PR and created a separate story for storybook update

src/hooks/useConsoleErrors.ts
```diff
@@ -2,23 +2,18 @@ import { useEffect } from 'react';
...
+    if (!callbacks?.onAddToCartClick) {
       // eslint-disable-next-line no-console
-      console.error('resultsPageOptions is a required field of type object');
+      console.error('callbacks?.onAddToCartClick is a required field of type function');
```
esezen, line 14: `?` is probably here because of copy paste. Can we change it to `callbacks.onAddToCartClick`?

src/stories/Quiz/argTypes.ts
```diff
@@ -1,32 +1,58 @@
...
   primaryColor: {
+    description: "RGB value string for quiz primary theme color ie: `'255, 82, 48'`",
     control: {
       type: 'text',
     },
   },
+  resultCardOptions: {
+    description: `Result card configuration options
+      (Required for the component interface and has to match the keys in the API response to be correctly rendered)
+      \`resultCardRegularPriceKey: string\`
```
esezen, line 38: Can we add a `\n` to this line to force it to go to the next line?

esezen (commented):
This is looking great. I like the new format a lot. Let's get another review from the team since this is a big PR and coordinate the release with the affected parties.

stanlp1 (changes requested):
The results page for "Render in a small container" "Provide CIO client instance", "Pass callbacks" are broken for me and don't render 🤔

mocca102:
> The results page for "Render in a small container" "Provide CIO client instance", "Pass callbacks" are broken for me and don't render 🤔

@stanlp1 resultCardOptions was missing. Passed it as a parameter 

Also we have this in the docs to indicate that it's required 
"Result card configuration options (Required for the component interface and has to match the keys in the API response to be correctly rendered)"

jjl014 (commented):
Nice, this is looking great to me!

I leave a few comments and have some questions.

The way that the description is currently set up is a bit hard to read as there's various text that are wrapped in backticks and the type that gets appended at the end. I created a [PR here](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/134) on a potential solution, but I'm not sure if it's a valid approach. 🤔 

Let me know what you think!

src/components/ResultContainer/ResultContainer.tsx
```diff
@@ -3,58 +3,63 @@ import RedoButton from '../RedoButton/RedoButton';
 import QuizContext from '../CioQuiz/context';
 import ResultFilters from '../ResultFilters/ResultFilters';
 import ZeroResults from '../ZeroResults/ZeroResults';
-import { ResultCardProps } from '../../types';
+import { ResultCardOptions } from '../../types';
 import Results from '../Results/Results';
 import Spinner from '../Spinner/Spinner';
 
 export interface IResultContainerProps {
```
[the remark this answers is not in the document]
mocca102: No particular reason the letter `I` stands for interface just to indicate that it's an interface. I just was not consistent

src/hooks/useQuizEvents/index.ts
```diff
@@ -26,10 +26,7 @@ const useQuizEvents: UseQuizEvents = (quizOptions, cioClient, quizState) => {
     quizLocalState,
     quizSessionStorageState,
   } = quizState;
-  const { resultsPageOptions, callbacks } = quizOptions;
-
-  const { onAddToCartClick, onQuizResultClick, onQuizResultsLoaded, onAddToFavoritesClick } =
-    resultsPageOptions;
+  const { callbacks } = quizOptions;
```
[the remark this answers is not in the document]
mocca102: Callback is optional so I couldn't destructure it without an If condition which meant doing let first for all of the callbacks so I figured accessing them like `callbacks?.[function]` was easier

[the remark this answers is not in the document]
jjl014: Ooo, got it. I think the following would work, but feel free to ignore this. 😄 

```
const { callbacks } = quizOptions || {};
const { onAddToCartClick, ... } = callbacks || {};
```

mocca102:
> Nice, this is looking great to me!
> 
> I left a few comments and a question.
> 
> Is there anyway to break the nested options out of the description? The way that the description is currently set up is a bit hard to read as there's various text that are wrapped in backticks and the type that gets appended at the end. I created a [PR here](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/134) on a potential solution, but I'm not sure if it's a valid approach. 🤔
> 
> Let me know what you think!
> 
> cc @esezen

@jjl014 I'm afraid this won't work. The approach in the draft you shared for this snippet for example 
```
 onAddToCartClick: {
    description: '...',
    table: {
      subcategory: 'callbacks',
    },
  }

```

Storybook treats `onAddToCartClick` as a direct prop to CioQuiz and this creates problems with type matching when we pass the argTypes to the story config since `onAddToCartClick` isn't a direct prop of CioQuiz but rather a prop on `callbacks`

@jjl014

src/hooks/useQuizEvents/index.ts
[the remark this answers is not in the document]
mocca102: nice trick

mocca102:
@jjl014 Merged your draft PR and added some tweak. Added callbacks.[function] instead of just function so it's clear to users that it's a property on callbacks. Let me know what you think about that.

Also hid the controls that aren't useful and left the ones we usually use.

src/stories/Quiz/Component/ComponentStories.stories.tsx
```diff
@@ -1,78 +1,50 @@
 /* eslint-disable no-console, react/jsx-props-no-spreading */
 import React, { useState } from 'react';
 import ConstructorIOClient from '@constructor-io/constructorio-client-javascript';
+import type { Meta } from '@storybook/react';
 import CioQuiz from '../../../components/CioQuiz';
-import { argTypes } from '../argTypes';
+import { argTypes, docsControls, storiesControls } from '../argTypes';
```
jjl014, line 6: Seems like we can remove `docsControl` from here as it's not used

src/stories/Quiz/Hooks/index.stories.tsx
```diff
@@ -3,7 +3,7 @@ import MyQuiz, { addHookStoryCode } from './index';
 import '../../../styles.css';
 import { basicDescription, apiKey, quizId, hookDescription } from '../../../constants';
 import { stringifyWithDefaults } from '../../../utils';
-import { argTypes } from '../argTypes';
+import { argTypes, docsControls, storiesControls } from '../argTypes';
 
 export default {
   title: 'Quiz/Hooks',
```
jjl014, line 9: Not sure if this is the right place, but it seems like the control for `primaryColor` isn't working here. Tried without quotes, with quotes, with single quotes. 🤔 

Also, why do some of the descriptions have a `-` and some don't? I think they were showing that the type was in the Component Basic Usage story, but it seems like those aren't shown here.

Screenshot 2023-10-05 at 5 25 20 PM

Also, this is just me being anal retentive, but could we be a bit more consistent with the capitalizations here?
* `import` => `Import`
* `use` => `Use`
* And I'll leave capitalizing the sub-bullets (i.e. `state management`) to your best judgement

Screenshot 2023-10-05 at 5 30 23 PM

src/stories/Quiz/Examples/index.stories.tsx
```diff
@@ -1,13 +1,16 @@
 /* eslint-disable no-console, react/jsx-props-no-spreading */
 import CioQuiz from '../../../components/CioQuiz';
-import { argTypes } from '../argTypes';
+import { argTypes, storiesControls } from '../argTypes';
 
 import RetrievingAnswersStory from './RetrievingAnswers';
 
 export default {
   title: 'Quiz/Specific Examples',
   component: CioQuiz,
```
jjl014, line 9: It looks like the controls on the `Specific Examples/Docs` page doesn't seem to be doing anything. Is it possible to fix them or should we disable the controls?

src/stories/Quiz/Component/ComponentStories.mdx
```diff
@@ -0,0 +1,126 @@
...
+- [Provide CIO Client Instance](#provide-cio-client-instance)
+- [Handle Favorites On Results Page](#handle-favorites-on-results-page)
+
+<h2>Quiz Props</h2>
+<ArgTypes of={e2eInteractionTest} />
+
+<Canvas>
+  <Story of={ComponentStories.BasicUsage} />
+</Canvas>
+<h2>Quiz Controls</h2>
```
jjl014, line 32: Would it make sense to change this to be `Example Quiz Controls` or `Sample Quiz Controls` since it only shows a subset of the possible config options?

src/stories/Quiz/Component/ComponentStories.mdx
```diff
@@ -0,0 +1,126 @@
...
+  );
+}
+``` 
+
+##### Usage Code Examples
+- [Basic Usage](#basic-usage)
+- [Provide CIO Client Instance](#provide-cio-client-instance)
+- [Handle Favorites On Results Page](#handle-favorites-on-results-page)
+
+<h2>Quiz Props</h2>
```
jjl014, line 26: I think a couple of the fields at the top do have defaults right?
* `enableHydration` - defaults to `false`?
* `primaryColor` - seems like this might have a default of a blue color? I could be wrong

src/stories/Quiz/tests/index.stories.tsx
```diff
@@ -20,33 +21,25 @@ export default {
...
-};
-
 const sessionStateOptions = {
   showSessionModal: false,
 };
 
 // eslint-disable-next-line storybook/prefer-pascal-case
 export const e2eInteractionTest = ComponentTemplate.bind({});
-e2eInteractionTest.args = { apiKey, quizId, resultsPageOptions, sessionStateOptions };
+e2eInteractionTest.args = {
```
jjl014, line 34: Hopefully, I'm leaving this comment in the right place.

Would it be possible to move the `Full Quiz e2e tests` story to the end? Also, moving Questions and Results right before it, but also towards the end.

Please reference
Screenshot 2023-10-05 at 5 32 45 PM
 my poor attached drawing.

jjl014 (commented):
Nice, I'm digging these changes. Thanks for making the updates!

I left a couple of comments. Also, is there anything we added previously that can be removed now that we're going with this approach?

Would also love to get some additional eyes from the @Constructor-io/customer-integrations team and get their thoughts as well.

Thanks!

src/stories/Quiz/Hooks/index.stories.tsx
mocca102: > Not sure if this is the right place, but it seems like the control for primaryColor isn't working here
The primary colors were not used in the hooks example. I added it in the example but it's not very user friendly.

> Also, why do some of the descriptions have a - and some don't? I think they were showing that the type was in the Component Basic Usage story, but it seems like those aren't shown here.
We have another PR for the hooks docs. But removed the `-` for now

src/stories/Quiz/Examples/index.stories.tsx
mocca102: Better to be removed

src/stories/Quiz/Component/ComponentStories.mdx
mocca102: In storybook Controls means options you configured to be controlled not necessarily all config options. I named it to just controls

src/stories/Quiz/Examples/index.stories.tsx
```diff
@@ -1,16 +1,8 @@
...
-import { argTypes, storiesControls } from '../argTypes';
-
 import RetrievingAnswersStory from './RetrievingAnswers';
 
 export default {
   title: 'Quiz/Specific Examples',
-  component: CioQuiz,
-  argTypes,
-  parameters: {
-    controls: storiesControls,
```
mocca102: Removed unnecessary controls

src/stories/Quiz/Hooks/index.stories.tsx
```diff
@@ -15,21 +21,23 @@ export default {
       description: {
         component: hookDescription,
       },
-      controls: docsControls,
```
mocca102: This is temporary until we push the hooks docs

src/stories/Quiz/Hooks/index.tsx
```diff
@@ -106,6 +107,7 @@ export default function HooksTemplate(args) {
       if (currentQuestion.isOpenQuestion) {
         return (
           <div className='cio-quiz'>
+            <style>.cio-quiz {convertPrimaryColorsToString(primaryColorStyles)}</style>
```
mocca102, line 110: `primaryColorStyles` wasn't being used in the hooks example. But this isn't a friendly customer interface so we still need to handle this in a separate PR maybe

jjl014 (approved):
Love all the changes. Thank you for putting up with my asks. ❤️ This look fantastic to me and I can't wait to get these changes out! 🚀 

I would love to get one other review from the team, but otherwise this looks amazing! Thanks again!

esezen (approved):
Love these changes! Thanks for working on this @mocca102 and thanks for the reviews @stanlp1 @jjl014 🚀 

I think we are good to merge this in, let's coordinate the `callback` changes with the affected parties to make sure we don't cause any issues

PR #128, opened 2023-09-12
Add support for favoriteItems array

mocca102:
Consumers of the library will need to provide `favoriteItems` array of Ids to handle favorites rendering state and whether to send conversion events or not.

### DIFF
--- a/src/hooks/usePropsGetters/index.ts
+++ b/src/hooks/usePropsGetters/index.ts
@@ -28,7 +28,8 @@ import useAddToFavoritesButtonProps from './useAddToFavoritesButtonProps';
 const usePropsGetters = (
   quizEvents: QuizEventsReturn,
   quizApiState: QuizAPIReducerState,
-  quizLocalState: QuizLocalReducerState
+  quizLocalState: QuizLocalReducerState,
+  favoriteItems?: string[]
 ) => {
   const {
     quizAnswerChanged,
@@ -96,8 +97,10 @@ const usePropsGetters = (
     [addToCart]
   );
 
-  const getAddToFavoritesButtonProps: GetAddToFavoritesButtonProps =
-    useAddToFavoritesButtonProps(addToFavorites);
+  const getAddToFavoritesButtonProps: GetAddToFavoritesButtonProps = useAddToFavoritesButtonProps(
+    addToFavorites,
+    favoriteItems
+  );
 
   const quizResultClickDown = useCallback(
     (event: KeyboardEvent<HTMLElement>, result: Partial<QuizResultData>, position: number) => {
--- a/src/hooks/usePropsGetters/useAddToFavoritesButtonProps.ts
+++ b/src/hooks/usePropsGetters/useAddToFavoritesButtonProps.ts
@@ -1,9 +1,11 @@
 import { useState, useCallback } from 'react';
+import { QuizResultData } from '@constructor-io/constructorio-client-javascript/lib/types';
 import { GetAddToFavoritesButtonProps, QuizEventsReturn } from '../../types';
 import { Favorited } from '../../components/ResultFavoritesButton/ResultFavoritesButton';
 
 export default function useAddToFavoritesButtonProps(
-  addToFavorites: QuizEventsReturn.AddToFavorites
+  addToFavorites: QuizEventsReturn.AddToFavorites,
+  favoriteItems?: string[]
 ): GetAddToFavoritesButtonProps {
   const [favorited, setFavorited] = useState<Favorited>({});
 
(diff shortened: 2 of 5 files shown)


src/hooks/usePropsGetters/useAddToFavoritesButtonProps.ts
```diff
@@ -14,13 +16,16 @@ export default function useAddToFavoritesButtonProps(
...
   );
 
+  const isResultFavorited = (result: Partial<QuizResultData>) =>
+    result?.data?.id && favoriteItems?.includes(result?.data?.id);
+
   const getAddToFavoritesButtonProps: GetAddToFavoritesButtonProps = useCallback(
     (result, price) => ({
       className: `${'cio-result-card-favorites-button'} ${
-        favorited?.[result?.data?.id || '0'] ? 'favorited' : ''
+        result?.data?.id && favoriteItems?.includes(result?.data?.id) ? 'favorited' : ''
```
esezen, line 25: If I'm understanding this correctly, the button's state doesn't depend on the local `favorited` state anymore. Since we are giving the responsibility of that to the consumer and using the passed `favoriteItems` for that purpose, can we get rid of `const [favorited, setFavorited] = useState<Favorited>({});`?

src/hooks/usePropsGetters/useAddToFavoritesButtonProps.ts
```diff
@@ -1,9 +1,11 @@
 import { useState, useCallback } from 'react';
+import { QuizResultData } from '@constructor-io/constructorio-client-javascript/lib/types';
 import { GetAddToFavoritesButtonProps, QuizEventsReturn } from '../../types';
 import { Favorited } from '../../components/ResultFavoritesButton/ResultFavoritesButton';
 
 export default function useAddToFavoritesButtonProps(
-  addToFavorites: QuizEventsReturn.AddToFavorites
+  addToFavorites: QuizEventsReturn.AddToFavorites,
+  favoriteItems?: string[]
```
esezen, line 8: I personally do not have a preference and I don't know which one is the more grammatically correct, but we had `favorited` vs `favorite` (new). Both of them sound correct to me tbh

src/hooks/usePropsGetters/useAddToFavoritesButtonProps.ts
```diff
@@ -14,13 +16,16 @@ export default function useAddToFavoritesButtonProps(
     [favorited]
   );
 
+  const isResultFavorited = (result: Partial<QuizResultData>) =>
+    result?.data?.id && favoriteItems?.includes(result?.data?.id);
+
   const getAddToFavoritesButtonProps: GetAddToFavoritesButtonProps = useCallback(
     (result, price) => ({
       className: `${'cio-result-card-favorites-button'} ${
```
esezen, line 24: Completely unrelated to this PR but I noticed while reviewing this, we only change the border color of the heart icon, would it make sense to fill it? I think the current one is hard to notice

esezen (commented):
Looking great. Left some comments

esezen (changes requested):
This is looking good. There is one thing we are missing and I don't know if we can support it in the story. It would be great if it's possible. When you click on "add to fav" in the component story, can we actually update the list and pass it back to the component? (Not talking about handling this in the hook, can we handle it in the story).
Again, not sure if this is possible to do. It would be great if we can do it because it looks "broken" right now since "favorited" doesn't get updated

esezen (approved):
LGTM!

esezen:
@mocca102 one more change, I just realized we don't have documentation for favoriteItems. Do you mind adding an explanation for it?

src/hooks/usePropsGetters/index.ts
```diff
@@ -137,17 +137,17 @@ const usePropsGetters = (
...
     }),
     [quizApiState.quizCurrentQuestion]
   );
 
   const getSelectQuestionImageProps: GetSelectQuestionImageProps = useCallback(
     (option) => ({
       className: 'cio-question-option-image',
-      src: option?.images?.primary_url,
-      alt: option?.images?.primary_alt,
+      src: option?.images?.primary_url || undefined,
```
mocca102, line 149: src and alt on img html element accept only type string or undefined and not null

src/stories/Quiz/Component/index.stories.tsx
```diff
@@ -151,3 +154,16 @@ addComponentStoryDescription(
   `const args = ${stringifyWithDefaults(PassCallbacks.args)}`,
   callbacksDescription
 );
+
```
mocca102, line 157: This isn't the best example but we're refactoring all of this on the new docs

esezen (approved):
LGTM!

PR #129, opened 2023-09-18
[csl-2811]: add answer option text to answerInputs

HHHindawy:
**This PR adds more details to the `answerInputs` array**

_Notice `answerInputs[5].value` & `answerInputs[6].value` -> Used to be an array of `string` now it's an array of objects consisting of `id` & `value`:_

**Before:**

image

**After:**

image


src/types.ts
```diff
@@ -96,7 +96,7 @@ export interface QuizReturnState {
 
 export type AnswerInput = {
   type: InputQuestionsTypes;
-  value: string | string[];
+  value: string | Partial<QuestionOption>[];
```
mocca102, line 99: Will this ever have any other props other than id and value? If not can we use Omit<> instead and omit the unused props?

src/components/CioQuiz/quizLocalReducer.ts
```diff
@@ -76,10 +76,12 @@ export default function quizLocalReducer(
...
           break;
         case QuestionTypes.SingleSelect:
-          newAnswers.push(currentAnswerInput.value as string[]);
-          break;
         case QuestionTypes.MultipleSelect:
-          newAnswers.push(currentAnswerInput.value as string[]);
+          newAnswers.push(
+            (typeof currentAnswerInput.value === 'string'
+              ? [currentAnswerInput.value]
+              : currentAnswerInput.value.map((answer) => answer.id)) as string[]
```
mocca102, line 83: I think if we did the Omit thing id will be required and you won't have to use `as string[]` here

mocca102 (changes requested):
This is looking good just left a couple of comments for typescript

src/types.ts
HHHindawy: True 👍

mocca102 (approved):

jjl014 (commented):
Nice work, this is looking great to me! 🚀 

I would love to get some thoughts on the comment I had below. If it sounds good, I think it would be an easy change to incorporate into this PR (but lmk otherwise). 😄

src/components/CioQuiz/quizLocalReducer.ts
```diff
@@ -76,10 +76,12 @@ export default function quizLocalReducer(
...
         case QuestionTypes.SingleSelect:
-          newAnswers.push(currentAnswerInput.value as string[]);
-          break;
         case QuestionTypes.MultipleSelect:
-          newAnswers.push(currentAnswerInput.value as string[]);
+          newAnswers.push(
+            (currentAnswerInput.value as Omit<QuestionOption, 'attribute' | 'images'>[]).map(
+              (answer) => answer.id
+            )
+          );
```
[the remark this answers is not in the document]
jjl014: I just realized the Skip code wasn't added by us. @HHHindawy, feel free to ignore this.

This PR looks great to me and I'll merge/ship it sometime today!

PR #131, opened 2023-09-21
[csl-2853]: include package version in c parameter

HHHindawy:


esezen (approved):
Tested in Storybook and by pulling the bundle to a code sandbox. Both seem to be working as expected. LGTM!

PR #145, opened 2023-10-19
[QUIZ-602] Copy quiz results and send email UI

lws803:
Adds copy quiz results functionality

https://github.com/Constructor-io/constructorio-ui-quizzes/assets/7874219/cb492d59-3954-44c4-ba36-8c743e03a465


lws803:
@mocca102 I wasn't able to pass `onEmailResults` callback to storybook and e2e tests though unfortunately. Will need some time to look into it but was wondering if you know what the issue is right away 🙂

lws803:
> @mocca102 I wasn't able to pass `onEmailResults` callback to storybook and e2e tests though unfortunately. Will need some time to look into it but was wondering if you know what the issue is right away 🙂

Oh nvm 👍 figured it out here: 4e8c041e63ee0b808f2d79b504b85cadb02b1ecb

src/types.ts
```diff
@@ -205,6 +209,12 @@ export interface ResetQuizButtonProps {
   style?: Record<string, string>;
 }
 
+export interface ShareResultsButtonProps {
+  className: string;
+  onClick?: React.MouseEventHandler<HTMLElement>;
+  style?: Record<string, string>;
```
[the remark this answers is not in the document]
lws803: Good call 👍 4e763951c8fb0f85f2587d554ace5ebc47c72f1f

v-i-s-h-n-u-ps (approved):
@lws803 Thank you. LGTM!

package.json
```diff
@@ -102,5 +102,11 @@
     "vite": "^4.2.1",
     "vite-plugin-css-injected-by-js": "^3.1.0",
     "webpack": "^5.75.0"
+  },
+  "dependencies": {
+    "@hookform/resolvers": "^3.3.2",
+    "react-hook-form": "^7.47.0",
+    "react-responsive-modal": "^6.4.2",
+    "zod": "^3.22.4"
```
mocca102, line 110: Sharing my comments about adding these dependencies here as well. We usually try to keep the dependencies few as possible to keep a small size for the bundled version of our library as well. I believe some dependencies might be convenient but some are unnecessary also and don't save us much dev time. I'm tagging @esezen here as well to see what he thinks. 

Sharing gzipped sizes here for easier lookup
"@hookform/resolvers": "923b",
"react-hook-form": "7.4k",
"react-responsive-modal": "3.8k",
"zod": "13.5"

React is 2.7k gzipped for comparison

src/components/ShareResultsModal/ShareResultsModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+    quizState.results?.response?.results
+      .filter((item) => item.data?.id)
+      .map((item) => item.data!.id)
+      .join(',') || ''
+  );
+  existingParams.set(
+    'attributes',
+    quizState.selectedOptionsWithAttributes?.map((option) => option).join(',') || ''
+  );
+  const value = urlObj.toString();
```
mocca102, line 31: Can we make this functionality available for the hooks interface as well? Like if people are using the hooks interface they won't have access to this feature. Depending on how quick we want this out I will leave it to you if you agree with me to decide whether to refactor this in this PR to be into a hook (something like useShareResultsLink() or do it in another PR.

src/stories/Quiz/argTypes.ts
```diff
@@ -82,6 +82,20 @@ export const argTypes = {
       },
     },
   },
+  'callbacks.onEmailResults': {
+    description: 'Callback function to be called when emailing results./',
```
mocca102, line 86: I believe the `/` at end of line here is extra

src/components/ShareResultsModal/EmailField.tsx
```diff
@@ -0,0 +1,56 @@
...
+          </div>
+          <button className='cio-share-results-share-button' type='submit'>
+            Send
+          </button>
+        </div>
+      </form>
+      {isSubmitSuccessful && (
+        <div className='cio-share-results-copied-notification'>
+          <CheckMarkCircleSVG />
+          <div>Email sent</div>
```
mocca102, line 47: I can imagine this becoming a problem since it indicates that the email has been sent when it actually means that the form has been submitted. Not sure how to handle this though

mocca102:
Are we going to add the handling of the loading of the copied url and rendering the results in a separate PR? @lws803

mocca102 (commented):
@lws803 Thanks for working on that awesome work! My only comments are about the used dependencies and refactoring that can wait if we have to get this out sooner.

Will request review from @esezen to see what he thinks

lws803:
> Are we going to add the handling of the loading of the copied url and rendering the results in a separate PR? @lws803

Yes there will be a separate PR for that 😄 https://github.com/Constructor-io/constructorio-ui-quizzes/pull/146

src/components/ShareResultsModal/EmailField.tsx
lws803: Yeah this is a weird one :x I think what we can do to make this better is to make email submit callback async and only show this message if the callback successfully resolves

lws803: ab384b91031ea7727d4bf0bda8d162af1a9b9e6e handleSubmit from react-hook-forms takes an async `onSubmit` function and only changes state to `isSubmitSuccessful` once the promise resolves

src/components/ShareResultsModal/ShareResultsModal.tsx
lws803: Oh good call! That won't be hard to move 👍 d280bd3f39a75f56bc6e87fe8e9363857411cf1e let me know if this is what you intended though. Also, do I have to update documentation or ts docs?

src/stories/Quiz/argTypes.ts
lws803: 030dc0f3119ecfe746210561c897aa72e60855db

src/components/ShareResultsModal/ShareResultsModal.tsx
lws803: Oh good call! That won't be hard to move 👍 d280bd3f39a75f56bc6e87fe8e9363857411cf1e let me know if this is what you intended though. Also, do I have to update documentation or ts docs?

package.json
lws803: I'm still leaning on the side of keeping react-hook-forms but maybe remove zod and hookform resolvers 😛 react-hook-forms helps us with state management which would otherwise be a little more work to track and implement.

src/components/ShareResultsModal/ShareResultsModal.tsx
mocca102: Thanks for adding this. We will follow up with a usage example after the new docs PR gets merged

src/components/ShareResultsModal/EmailField.tsx
mocca102: Nice 🧠

src/components/ShareResultsModal/ShareResultsModal.tsx
mocca102: We will follow up with a usage example after the new docs PR gets merged

mocca102: Thanks for refactoring

mocca102:
Thanks for making the changes!

package.json
esezen: I'm with @mocca102 on this one. With all of our libraries, we take dependencies very seriously and try to have as few of them as possible. Customers also value this highly and even with [client-js](https://github.com/Constructor-io/constructorio-client-javascript) which has only a handful (we are removing 1 dependency soon and 1 in the mid-term), we heard concerns from customers about size. Right now this PR is getting the bundle size above 100KB with a 30% increase in size. 

I believe `@hookform/resolvers` and `zod` are for validation. Considering the amount of validation we do and the complexity, I think we can get rid of these. What do you think? For `react-hook-form`, I'm still in the middle of reviewing the PR so I'm not entirely sure about how it's used. If the work needed to remove it and have our own implementation is around a day's work, I would vote to have our implementation as well. If you think it's more, let's move this to Slack or a to a meeting and discuss internally. There are concerns with dependencies other than size as well.

src/components/ShareResultsModal/EmailField.tsx
```diff
@@ -0,0 +1,56 @@
...
+        </div>
+      )}
+    </div>
+  );
+}
+
+type Props = {
```
esezen, line 54: Not a big deal but it would be better if we follow the existing patterns. We usually put the types to the top of the file as an interface. Can we do that just to keep it consistent? Tagging TS wizard @mocca102 

https://github.com/Constructor-io/constructorio-ui-quizzes/blob/bedaf3e82e855237e2dcef1854a1846c6864a82b/src/components/RedoButton/RedoButton.tsx#L5

src/components/ShareResultsModal/LinkField.tsx
```diff
@@ -0,0 +1,35 @@
...
+        <div className='cio-share-results-copied-notification'>
+          <CheckMarkCircleSVG />
+          <div>Link copied to clipboard</div>
+        </div>
+      )}
+    </div>
+  );
+}
+
+type Props = {
```
esezen, line 33: Same as above

src/components/ShareResultsModal/ShareResultsModal.tsx
```diff
@@ -0,0 +1,61 @@
...
+            : 'Share or save your quiz results with this link.'}
+        </div>
+        {onEmailResults && <EmailField onSubmit={(email) => onEmailResults({ email, url })} />}
+        <LinkField url={url} />
+      </div>
+    </Modal>
+  );
+}
+
+type Props = {
```
esezen, line 55: Same as above

src/components/ShareResultsModal/EmailField.tsx
```diff
@@ -0,0 +1,56 @@
...
+                {errors.email.message}
+              </div>
+            )}
+          </div>
+          <button className='cio-share-results-share-button' type='submit'>
+            Send
+          </button>
+        </div>
+      </form>
+      {isSubmitSuccessful && (
```
esezen, line 44: If `onEmailResults` callback throws an error, we don't display anything on the page. Is this expected? Do we want the enduser to handle that case? (I'm not against it, just checking)

src/components/ShareResultsModal/LinkField.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <div className='cio-share-results-feature-group'>
+      <div className='cio-share-results-description'>Share by link</div>
+      <div className='cio-share-results-button-group'>
+        <input className='cio-share-results-link-input' value={url} disabled />
+        <button
+          className='cio-share-results-share-button'
+          type='button'
+          onClick={() => {
+            navigator.clipboard.writeText(url);
```
esezen, line 17: I believe this has the potential to throw an error. Can we wrap it in try catch?

package.json
lws803: Oh yeah for sure, I can remove the validator libraries 👍 agreed it makes no sense to blow up our bundle size if we're just using it for one validation (email). I'm on the fence with react-hook-form though since it is pretty helpful in structuring code and maintaining states. It's definitely something I can replace with our own within a day as well but it's nice to have that library manage form states for us 🙂

src/components/ShareResultsModal/ShareResultsModal.tsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import Modal from 'react-responsive-modal';
```
esezen, line 2: What are the benefits of this library? We already have a modal here: https://github.com/Constructor-io/constructorio-ui-quizzes/blob/main/src/components/SessionPromptModal/SessionPromptModal.tsx

Does this library handle some fancy stuff?

esezen (commented):
This is great! Thanks for working on this 🚀 

Left some comments

src/components/ShareResultsModal/EmailField.tsx
lws803: That's a good question, there are no designs for the error state unfortunately 🤔 but I'll check with Mariano in case he has some thoughts on this.

src/components/ShareResultsModal/ShareResultsModal.tsx
lws803: Mostly just modal placement and also handling clicks outside of the modal itself 😄 I can remove this though if you feel strongly about not using this package. Sorry, I should've checked first on the strict requirement to keep bundle sizes small 😅

src/components/ShareResultsModal/EmailField.tsx
lws803: Whoops, good catch! Sorry, I should've aligned with the existing patterns instead of borrowing practices from another 😛 27f85b661c45a8c98ce347cd87ae97e484c8900e

lws803: Whoops, good catch! Sorry, I should've aligned with the existing patterns instead of borrowing practices from another 😛 27f85b661c45a8c98ce347cd87ae97e484c8900e

lws803: Whoops, good catch! Sorry, I should've aligned with the existing patterns instead of borrowing practices from another 😛 75b49eb303d05807142008881e97dc49fef9d339

src/components/ShareResultsModal/LinkField.tsx
lws803: 75b49eb303d05807142008881e97dc49fef9d339

src/components/ShareResultsModal/ShareResultsModal.tsx
lws803: 75b49eb303d05807142008881e97dc49fef9d339

src/hooks/useQuizState/useUrlState.ts
```diff
@@ -0,0 +1,4 @@
+export default function useUrlState() {
```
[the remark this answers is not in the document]
lws803: Removed it here 75b49eb303d05807142008881e97dc49fef9d339

package.json
lws803: Removed the validators and resolvers here 👍 2d8e9c5387b458dbf10eb8f13ff7e0cff41b364f

src/components/ResultContainer/ResultContainer.tsx
```diff
@@ -31,9 +33,12 @@ export default function ResultContainer(props: IResultContainerProps) {
       return (
         <div className='cio-results-container'>
           <h1 className='cio-results-title'>{resultsTitle}</h1>
-          <div className='cio-results-filter-and-redo-container'>
+          <div className='cio-results-button-group'>
             <ResultFilters hasNoResults={zeroResults} />
-            <RedoButton />
+            <div className='cio-results-redo-and-share-button-group'>
+              <RedoButton />
+              <ShareButton onClick={onShare} />
```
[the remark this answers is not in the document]
lws803: That's a good question 🤔 let me check up on that as well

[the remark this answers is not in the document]
lws803: Added a new prop here 34282f1773f8abcad715fa444f8911fcf63db8f1

src/stories/Quiz/argTypes.ts
```diff
@@ -82,6 +82,20 @@ export const argTypes = {
       },
     },
   },
+  'callbacks.onEmailResults': {
```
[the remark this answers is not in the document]
lws803: Hmm that's a good suggestion, but I also wanna avoid providing more access unless they specifically ask for it though 🤔 just to keep things on the leaner side

src/components/ResultContainer/ResultContainer.tsx
```diff
@@ -31,9 +33,12 @@ export default function ResultContainer(props: IResultContainerProps) {
       return (
         <div className='cio-results-container'>
           <h1 className='cio-results-title'>{resultsTitle}</h1>
-          <div className='cio-results-filter-and-redo-container'>
```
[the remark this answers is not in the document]
lws803: Good call, 3ac51fa9ddcf71f28e5b9ef3891853b83390547b added both classes. `cio-results-filter-and-redo-container` will not technically exist in css but it can be used to preserve backwards compatibility in case folks are using it with selectors

src/components/ShareResultsModal/EmailField.tsx
lws803: CleanShot 2023-10-31 at 11 19 29@2x

Added an error state here 👍

src/components/ShareResultsModal/LinkField.tsx
lws803: CleanShot 2023-10-31 at 11 20 08@2x

Added an error for copying as well

lws803: 85a550b81a9df26541cac20a5d7d5af28d154ad3

src/components/ShareResultsModal/EmailField.tsx
lws803: 85a550b81a9df26541cac20a5d7d5af28d154ad3

src/components/ShareResultsModal/ShareResultsModal.tsx
lws803: Removed modal package 👍 ad9ec88

package.json
esezen: For `react-hook-form`, I still think it's better to implement our own instead of bringing a library to handle a single input but let me tag @jjl014 to see what he thinks

lws803: Ahh no worries, if you feel strongly about it, I can remove it as well 👍 it'll just be a few states to maintain on our end

jjl014: Hey Team, I just took a look and I agree with @mocca102 and @esezen here. Generally we try to keep our libraries as lean as possible as we've had complaints from customers before in regards to some of our other libraries. They can be pretty stingy when it comes to size and every byte counts unfortunately. 😓 

From what I can tell looking at the code, it looks like we're using the package for a small portion of the component which can be handled by simple `useState` hooks (but please correct me if I'm wrong). For a package that takes up such a large size of our library, it doesn't seem worth it and probably makes more sense to roll our own.

lws803: Makes sense to me as well 👍 have switched to using our own states. Thanks for clarifying @jjl014 @esezen @mocca102 !

esezen (approved):
LGTM!

PR #146, opened 2023-10-23
[QUIZ-603]: shared results page

gulevskydev:
For testing: 
Go to - https://codesandbox.io/s/quizzes-js-bundle-example-forked-cvhg3z
Insert URL: https://cvhg3z.csb.app/?attributes=red,blue,yellow&items=119011081
1) Should render results page
2) If clicked retake - should remove query params 'item' and 'attributes'  and start quiz again from first question

### DIFF
--- a/src/hooks/useQueryParams.ts
+++ b/src/hooks/useQueryParams.ts
@@ -1,0 +1,28 @@
+import { useCallback } from 'react';
+
+const useQueryParams = () => {
+  const queryParams = new URLSearchParams(window.location.search);
+
+  const getQueryParam = (queryName) => {
+    const queryParam = queryParams.get(queryName);
+    if (!queryParam) return [];
+
+    return queryParam?.split(',');
+  };
+
+  const removeQueryParams = useCallback(() => {
+    const updatedUrl = new URL(window.location.href);
+    updatedUrl.searchParams.delete('items');
+    updatedUrl.searchParams.delete('options');
+
+    window.history.replaceState({}, '', updatedUrl.toString());
+  }, []);
+
+  const queryItems = getQueryParam('items');
+  const queryOptions = getQueryParam('options');
+  const isSharedResultsQuery = !!queryItems.length && !!queryOptions.length;
+
+  return { queryItems, queryOptions, isSharedResultsQuery, removeQueryParams };
+};
+
+export default useQueryParams;
--- a/src/hooks/useQueryParams.ts
+++ b/src/hooks/useQueryParams.ts
@@ -1,31 +1,39 @@
 import { useCallback } from 'react';
 
 const useQueryParams = () => {
-  const getQueryParam = (queryName) => {
-    const queryParams = new URLSearchParams(window.location.search);
+  const queryParams = new URLSearchParams(window.location.search);
+  const getParsedQueryParam = (queryName) => {
     const queryParam = queryParams.get(queryName);
     if (!queryParam) return [];
 
     return queryParam?.split(',');
   };
 
+  const queryItems = getParsedQueryParam('items');
+  const queryAttributes = getParsedQueryParam('attributes');
+  const quizBasePath = queryParams.get('quizBasePath');
+  const isSharedResultsQuery = !!queryItems.length && !!queryAttributes.length;
+
   const removeSharedResultsQueryParams = useCallback(() => {
-    const updatedUrl = new URL(window.location.href);
+    const updatedUrl = new URL(quizBasePath ?? window.location.href);
     updatedUrl.searchParams.delete('items');
     updatedUrl.searchParams.delete('attributes');
+    updatedUrl.searchParams.delete('quizBasePath');
 
     if (!updatedUrl.searchParams.toString().length) {
       updatedUrl.search = '';
     }
 
     window.history.replaceState({}, '', updatedUrl.toString());
-  }, []);
+  }, [quizBasePath]);
 
-  const queryItems = getQueryParam('items');
-  const queryAttributes = getQueryParam('attributes');
-  const isSharedResultsQuery = !!queryItems.length && !!queryAttributes.length;
-
-  return { queryItems, queryAttributes, isSharedResultsQuery, removeSharedResultsQueryParams };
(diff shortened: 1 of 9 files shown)


src/hooks/useQueryParams.ts
```diff
@@ -0,0 +1,31 @@
...
+  const getQueryParam = (queryName) => {
+    const queryParams = new URLSearchParams(window.location.search);
+    const queryParam = queryParams.get(queryName);
+    if (!queryParam) return [];
+
+    return queryParam?.split(',');
+  };
+
+  const removeSharedResultsQueryParams = useCallback(() => {
+    const updatedUrl = new URL(window.location.href);
```
lws803, line 13: **suggestion**: I wonder if we should use `quizBasePath` prop like in my PR https://github.com/Constructor-io/constructorio-ui-quizzes/pull/145/files#diff-4ca8c4ece22292244e023432ad4ad692edfa1d8e0a64e44537935be42225251bR115 if supplied

lws803: but using window location href could be the fallback behavior though

src/hooks/useQuizState/useQuizApiState.ts
```diff
@@ -59,13 +60,29 @@ const useQuizApiState: UseQuizApiState = (
     }
   };
 
+  const dispatchSharedQuizResults = async () => {
+    try {
+      const quizResults = await getBrowseResultsForItemIds(cioClient, queryItems);
```
lws803, line 65: **comment**: That came in handy :)

lws803 (changes requested):
Looking good, really clean code. ✨  Tested it on the sandbox link you've shared, works like a charm!

Just one suggestion regarding considering the `quizBasePath` prop.

src/hooks/useQueryParams.ts
```diff
@@ -1,31 +1,39 @@
...
+  const getParsedQueryParam = (queryName) => {
     const queryParam = queryParams.get(queryName);
     if (!queryParam) return [];
 
     return queryParam?.split(',');
   };
 
+  const queryItems = getParsedQueryParam('items');
+  const queryAttributes = getParsedQueryParam('attributes');
+  const quizBasePath = queryParams.get('quizBasePath');
```
lws803, line 14: **issue**: Oh `quizBasePath` is not supplied as a query param here, it's supplied as a prop, see https://github.com/Constructor-io/constructorio-ui-quizzes/pull/145/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR88 you might have to copy some of my changes over though

lws803 (changes requested):

src/hooks/useQueryParams.ts
lws803: It will most likely be the same domain but just incase customers would like to pre-specify some query params, they can do that 👍 

> Because if we will try to update url with not the same domain we will get the error because of XSS

Ahh good shout, I don't think they'll specify any domain outside of their own 🤔

src/hooks/useQuizEvents/useQuizResetClick.ts
```diff
@@ -5,16 +5,23 @@ import {
...
 import useQueryParams from '../useQueryParams';
 
 const useQuizResetClick = (
   resetQuizSessionStorageState: () => void,
   dispatchLocalState: React.Dispatch<ActionAnswerQuestion>,
   dispatchApiState: React.Dispatch<ActionQuizAPI>,
-  quizResults?: QuizResultsResponse | QuizSharedResultsData
+  quizResults?: QuizResultsResponse | QuizSharedResultsData,
+  quizBasePath?: IQuizProps['quizBasePath']
+  // eslint-disable-next-line max-params
```
lws803, line 22: **suggestion**: I wonder if we should refactor this and pass an object as function parameter instead

src/hooks/useQueryParams.ts
lws803: As discussed, let's remove `quizBasePath` for now 👍 sorry for the trouble!

src/hooks/useQueryParams.ts
```diff
@@ -0,0 +1,38 @@
+import { useCallback } from 'react';
+
+import { IQuizProps } from '../types';
+
+const useQueryParams = (quizBasePath: IQuizProps['quizBasePath']) => {
```
lws803, line 5: **comment**: Let's revert `quizBasePath` param as discussed 😅

lws803 (changes requested):

lws803 (approved):
LGTM 👍 thanks for addressing the comments and sorry for the slight detour 😛 let's get a review from @mocca102 before merging though 👍

mocca102 (approved):
LGTM! The query parameters don't get removed on reset quiz when tested on storybook but might be smth only related to storybook. Also I was a bit concerned about rendering what's in the query params but React seems to be a good job escaping unwanted characters. Thanks for working on that!

mocca102:
Also just noticed that on this url [](https://cvhg3z.csb.app/?attributes=red,blue,yellow&items=119011081)

Clicking retake quiz throws an error, Might be an edge case we need to handle when there's `/?` @v-i-s-h-n-u-ps @lws803 
Screenshot 2023-10-27 at 4 28 48 AM

gulevskydev:
> https://cvhg3z.csb.app/?attributes=red,blue,yellow&items=119011081

Sorry, it was not the last version of compiled code in sandbox, now you can check the same logic using this URL: https://7q7w9x.csb.app/?attributes=red,blue,yellow&items=119011081

gulevskydev:
> Also just noticed that on this url
> 
> Clicking retake quiz throws an error, Might be an edge case we need to handle when there's `/?` @v-i-s-h-n-u-ps @lws803 Screenshot 2023-10-27 at 4 28 48 AM

Sorry, it was not the last version of compiled code in sandbox, now you can check the same logic using this URL: https://7q7w9x.csb.app/?attributes=red,blue,yellow&items=119011081

PR #149, opened 2023-11-06
[CSL-2974] Fix select question options images height

HHHindawy:


mocca102 (changes requested):
Image options looking good. The background image for select questions that we discussed still doesn't fill 100% of the height when the view width gets smaller 

Screenshot 2023-11-07 at 4 05 27 AM

HHHindawy:
> Image options looking good. The background image for select questions that we discussed still doesn't fill 100% of the height when the view width gets smaller

Are you sure you're on the right branch? @mocca102  🤔 

image

mocca102 (approved):
LGTM! lets get a second review. Just making it a habit of two reviewers for UI changes

HHHindawy:
Yes @esezen, this was intentional as agreed with @mocca102.
We wanted to fix the image ratio to unify how images look in case we get different sized images in a single question.

PS: The `aspect-ratio: 1/1` property was always there but didn't take correct effect as it was placed on the wrong element.

PR #155, opened 2023-11-29
[QUIZ-613] subtask integrate new fields os library

v-i-s-h-n-u-ps:


src/components/ResultContainer/ResultContainer.tsx
```diff
@@ -26,13 +26,36 @@ export default function ResultContainer(props: IResultContainerProps) {
     renderResultCardPriceDetails,
   } = resultCardOptions || {};
   const numberOfResults = state?.quiz.results?.response?.results?.length;
+  const resultsConfig = state?.quiz.resultsConfig?.desktop;
```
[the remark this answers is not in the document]
v-i-s-h-n-u-ps: Forgot why we agreed on this structure. But this value wil be used for all viewports.

src/components/ResultContainer/ResultContainer.tsx
```diff
@@ -26,13 +26,36 @@ export default function ResultContainer(props: IResultContainerProps) {
...
+    resultsDescription = resultsConfig.description.text ?? '';
+  } else {
+    resultsDescription = '';
+  }
 
   if (state?.quiz.results) {
     return (
       <div className='cio-results-container'>
         <h1 className='cio-results-title'>{resultsTitle}</h1>
+        <p className='cio-results-description'>{resultsDescription}</p>
```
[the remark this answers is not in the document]
v-i-s-h-n-u-ps: Updated styles for description.

src/components/ResultContainer/ResultContainer.tsx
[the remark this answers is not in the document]
esezen: Why is it named `desktop` if it's being used for all viewports? I think it will be okay for UI library users since they don't interact with it directly but for anyone doing an API integration, this will be confusing

src/hooks/useQuizState/useQuizApiState.ts
```diff
@@ -75,6 +90,13 @@ const useQuizApiState: UseQuizApiState = (
     }
   };
 
+  useEffect(() => {
```
esezen, line 93: I think this fixed the double fetch issue but introduced a new one. If you go through the quiz once, everything works fine. If you click on retake and take it again, the config is not fetched and on top of that the even the default title doesn't render.
Screenshot 2023-12-01 at 9 51 24 AM

esezen (changes requested):
Thanks for making the changes!
1. I left one comment about the `state.quiz.resultsConfig.desktop`
2. There seems to be a problem with fetching the config when you retake the quiz

src/hooks/useQuizState/useQuizApiState.ts
v-i-s-h-n-u-ps: Fixed here [17dee0c](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/155/commits/17dee0c0f773c9e17295bc829bdca75424594341)

src/components/ResultContainer/ResultContainer.tsx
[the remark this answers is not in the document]
v-i-s-h-n-u-ps: I believe the idea behind was that users in the future can have different texts for different viewports. For example a longer description for bigger screens and shorter one for mobile and tablets. But for an mvp, only the desktop implementation is available as of now.

src/hooks/useQuizState/useQuizApiState.ts
```diff
@@ -75,6 +90,13 @@ const useQuizApiState: UseQuizApiState = (
...
+    (async () => {
+      dispatchQuizResultsConfig();
+    })();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
```
mocca102, line 99: Do we need to wrap dispatchQuizResultsConfig in an async function here?

Since we're just calling it I believe we can just do right?
```
useEffect(() => {
  dispatchQuizResultsConfig();
}, []);
```

src/components/CioQuiz/quizApiReducer.ts
```diff
@@ -91,9 +93,15 @@ export default function apiReducer(
...
+    case QuizAPIActionTypes.SET_QUIZ_RESULTS_CONFIG: {
+      return {
+        ...state,
+        resultsConfig: action.payload?.quizResultsConfig.results_config,
+      };
+    }
 
     case QuizAPIActionTypes.RESET_QUIZ:
-      return initialState;
+      return { ...initialState, resultsConfig: state.resultsConfig };
```
esezen, line 104: Why are we not resetting `resultsConfig` isn't it possible it can change?

v-i-s-h-n-u-ps: [7a5dd35](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/155/commits/7a5dd35b63e66ffec87f0e794ff0d8b6f695793a)

src/hooks/useQuizState/useQuizApiState.ts
mocca102: @v-i-s-h-n-u-ps Sorry I didn't mention this from the start but looking further into this I'm thinking we need to call `dispatchQuizResultsConfig();` where we call `dispatchQuizResults();` instead of calling it in its own effect (like screenshot 1). This way both of them are called/updated together. The only problem I faced was that `getResultsConfig` get's called twice. The same problem happens with getSearchResults as well but we don't see the second API request since it throws an error (screenshot 2)

Can we call both dispatchQuizResultsConfig and dispatchQuizResults together and fix the problem that both are being called twice? I know the duplicate requests issue isn't related to this PR so please let me know If you need any help here.

Screenshot 2023-12-04 at 6 52 58 PM

----------------------------------

Screenshot 2023-12-04 at 6 55 41 PM
)

src/hooks/useQuizState/useQuizApiState.ts
```diff
@@ -60,6 +65,16 @@ const useQuizApiState: UseQuizApiState = (
     }
   };
 
+  const dispatchQuizResultsConfig = async () => {
+    const quizResultsConfig = await getQuizResultsConfig(cioClient, quizId, {
```
mocca102, line 69: Can we wrap this in a try catch please

mocca102 (changes requested):
This looks great thanks for working on this I left two comments let me know what you think please

src/hooks/useQuizState/useQuizApiState.ts
v-i-s-h-n-u-ps: As discussed offline, leaving it in a separate `useEffect`.

src/hooks/useQuizState/useQuizApiState.ts
v-i-s-h-n-u-ps: [67cbc4c](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/155/commits/67cbc4ca06d9703347e8657339d6cbc7db350f44)

v-i-s-h-n-u-ps:
> We're close to get this pushed, There's a failing test related to the results title since it changes now. Can we fix failing tests please based on if there are result configs or not?
> 
> Screenshot 2023-12-04 at 8 07 53 PM

[10a91e6](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/155/commits/10a91e6341f7929e49f0bbc16d318734f6a97f12)

mocca102 (approved):
LGTM!

esezen (changes requested):
For quizzes that don't have any config set, this introduces a breaking change to the results page title
image

v-i-s-h-n-u-ps:
> 960d4eb

[960d4eb](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/155/commits/960d4eb17417410987b6dcc0463121f046398c8e)

esezen (approved):
LGTM!

PR #159, opened 2024-01-09
[CSL-3139] Fix styles in useQuiz live example

Mudaafi:
Styles weren't working in the Hooks Live Example not because of CSS but because we started using a new css class `cio-question-buttons-group` in the Components, which was missing in the example.

I also moved the "Control Bar" to be a top level .cio-quiz child to emulate our current style of having the control bar fill the whole container and be fixed to the bottom.

### DIFF
--- a/src/stories/Quiz/Hooks/index.tsx
+++ b/src/stories/Quiz/Hooks/index.tsx
@@ -108,18 +108,19 @@ export default function HooksTemplate(args) {
                   </div>
                 ))}
             </div>
-
-            {zeroResults && (
-              <div className='cio-zero-results'>
-                <h3 className='cio-zero-results-subtitle'>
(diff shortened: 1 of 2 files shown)


esezen (changes requested):
This is looking good. Thanks for fixing this. Would it be possible to have this border behave like others so it doesn't move around? 
border

Edit: I thought this behavior was introduced in this PR but I just realized it's the same on main. Let me know if you think it should be addressed in a separate PR and I can approve this

Mudaafi:
Nice eyes, I'll make the change in this PR since I didn't do as much css changes as expected

esezen (approved):
LGTM!

Mudaafi:
Added a transition property as well to match the `skip` button experience as well.

PR #169, opened 2024-01-29
Handle answers required for createQuizUrl function

mocca102:
This PR only wraps a code block with an if condition.

`dispatchQuizResults` was throwing an error when answers is an empty array (initially)

Added a condition to only getQuizResults when answers isn't empty.

### DIFF
--- a/src/hooks/useQuizState/useQuizApiState.ts
+++ b/src/hooks/useQuizState/useQuizApiState.ts
@@ -111,7 +111,7 @@ const useQuizApiState: UseQuizApiState = (
         await dispatchSharedQuizResults();
       } else if (skipToResults && quizLocalState.answers.length) {
         await dispatchQuizResults();
-      } else {
+      } else if (!skipToResults) {
         try {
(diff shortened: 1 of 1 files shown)


src/hooks/useQuizState/useQuizApiState.ts
```diff
@@ -40,28 +40,30 @@ const useQuizApiState: UseQuizApiState = (
...
+          type: QuizAPIActionTypes.SET_QUIZ_RESULTS,
+          payload: {
+            quizResults,
+          },
+        });
+        if (!quizLocalState.isQuizCompleted)
+          dispatchLocalState({
+            type: QuestionTypes.Complete,
+          });
+      } catch (error) {
```
mocca102, line 62: To test remove `if(quizLocalState.answers.length)` and add a debugger here. You'll see that for one of the rerenders there's an error for "answers required"

When we add a UI for the error state this will show an error message for a split second

The "if" prevents that

stanlp1 (approved):
Change makes sense, looks good to me.

esezen (approved):
LGTM!

PR #174, opened 2024-02-21
ability to configure fmt options

v-i-s-h-n-u-ps:


esezen:
@mocca102 to add `requestConfigs` to `ResultsPageOptions`

stanlp1 (approved):
Changes look good to me. 

Noticed a couple of things. Not sure if we need to address them in this story.

1. We are listing defaults for price keys, ratings, etc. But we don't actually have defaults.
2. Should we add a note about using dot notation for the keys?
3. Should we add a safety for when the key results in an object? Right now it breaks. 🤔

src/types.ts
```diff
@@ -50,10 +52,16 @@ export namespace QuizResultsEventsProps {
   export type OnEmailResults = (data: QuizEmailResults) => Promise<void>;
 }
 
+export type QuizResultsRequestConfigs = Omit<
+  QuizzesResultsParameters,
```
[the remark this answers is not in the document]
stanlp1: I think you might not have npm i'ed the update. Since it's in the new version of client js 🤔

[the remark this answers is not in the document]
Mudaafi: you're right, ignore my comment

src/components/ResultCard/ResultCard.tsx
```diff
@@ -90,9 +100,7 @@ export default function ResultCard(props: ResultCardOptions) {
 
   return (
     <div className='cio-result-card'>
-      {customAddToFavoritesCallback && (
-        <ResultFavoritesButton item={result} price={salePrice || regularPrice} />
-      )}
+      {customAddToFavoritesCallback && <ResultFavoritesButton item={result} price={salePrice} />}
```
esezen, line 103: @mocca102 It looks like we removed regularPrice from here while testing this out, can we bring it back?

src/components/ResultCard/ResultCard.tsx
```diff
@@ -44,20 +54,20 @@ export default function ResultCard(props: ResultCardOptions) {
         <p className='cio-result-card-title'>{result.value}</p>
         <div className='cio-result-card-details'>
           <div className='cio-result-card-rating'>
-            {ratingScore && (
+            {!!ratingScore && (
```
v-i-s-h-n-u-ps, line 57: Do any customer allow a `0` rating score? If so should we explicitly mention `ratingScore` to be defined

src/components/ResultCard/ResultCard.tsx
```diff
@@ -44,20 +54,20 @@ export default function ResultCard(props: ResultCardOptions) {
...
               <span className='cio-result-card-rating-score'>
                 {Array(Number(ratingScore)).fill('★')}
                 {Array(5 - Number(ratingScore)).fill('☆')}
               </span>
             )}
-            {ratingCount && <span className='cio-result-card-rating-count'>({ratingCount})</span>}
+            {!!ratingCount && <span className='cio-result-card-rating-count'>({ratingCount})</span>}
```
v-i-s-h-n-u-ps, line 63: Same concern here.

v-i-s-h-n-u-ps (commented):
The changes look mostly good to me. Left a couple of concerns that might be an issue.

src/components/ResultCard/ResultCard.tsx
esezen: Good callout. We discussed this and decided we want to hide these if the score is 0

PR #175, opened 2024-03-06
[QUIZ-689] post mortem quizzes os UI library test coverage set up

v-i-s-h-n-u-ps:


mocca102 (commented):
@v-i-s-h-n-u-ps Will we be adding server side rendering tests? I know we only support client side but many people are using remix/next recently which means the code have a lifecycle where it runs at least once on the server, Don't we want to make sure the the code just doesn't explode when this happen?

v-i-s-h-n-u-ps:
> @v-i-s-h-n-u-ps Will we be adding server side rendering tests? I know we only support client side but many people are using remix/next recently which means the code have a lifecycle where it runs at least once on the server, Don't we want to make sure the the code just doesn't explode when this happen?

Yes, we should. I'll see if there's any extra set up that needs to be done for server side tests. The purpose of this PR is only to set this repo up with RTL. Utility functions and helpers for both client and server will follow in another PR so as not to blow this one up.

.eslintrc.json
```diff
@@ -52,22 +52,35 @@
...
+      "plugins": ["jest", "no-snapshot-testing", "testing-library"],
+      "extends": [
+        "airbnb-typescript",
+        "prettier",
+        "plugin:@cspell/recommended",
+        "plugin:jest/recommended",
+        "plugin:testing-library/react"
+      ],
+      "rules": {
+        "testing-library/no-node-access": ["error", { "allowContainerFirstChild": true }],
```
mocca102, line 77: this is optional but I'd like if we added links to some of the eslint rule for others to know what they mean would be great.

I grabbed the link for you already 
- https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-node-access.md

src/components/SkipButton/SkipButton.test.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
```
mocca102, line 1: We have discussed before (CI) to keep tests out of src folder in their specs folded.

What do you think about that? Would like to know your opinion @v-i-s-h-n-u-ps

src/components/SkipButton/SkipButton.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { render, screen } from '@testing-library/react';
+
+import SkipButton from './SkipButton';
+
+describe(SkipButton, () => {
```
mocca102, line 6: How the title here is being used is the component itself and not a string? SkipButton

mocca102 (commented):
How to run the tests I'm getting this @v-i-s-h-n-u-ps 
Screenshot 2024-03-06 at 9 41 50 PM

src/components/SkipButton/SkipButton.test.tsx
v-i-s-h-n-u-ps: Jest is smart enough to pick up the name from the function or class unless they are memoized

.eslintrc.json
v-i-s-h-n-u-ps: [d4ee190](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/175/commits/d4ee1905b6eee8d21815a55f2716817ea09ce939)

src/components/SkipButton/SkipButton.server.test.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { renderToString } from 'react-dom/server';
+import SkipButton from './SkipButton';
+
+describe(SkipButton, () => {
```
mocca102, line 5: Can we add any prefix/suffix to the name that indicates that it's a server or a client test please?

mocca102 (approved):
I love this! Thanks for making the changes. LGTM!

Let's have a second quick review from @esezen and I think this is good to go 

Also on the 85% coverage not sure how we derived the number but lets see what Enes think as well

src/components/SkipButton/SkipButton.test.tsx
mocca102: I don't have strong preferences. @esezen what do you think?

v-i-s-h-n-u-ps:
>Also on the 85% coverage not sure how we derived the number but lets see what Enes think as well

No logic behind, just a random number that felt good enough. We can change it depending on how everyone feels about it

v-i-s-h-n-u-ps:
@mocca102 I've added some utility functions to the PR.

.eslintrc.json
```diff
@@ -52,22 +52,40 @@
...
-      "files": "tests/**/*.ts",
+      "files": ["**/*.test.{ts,tsx,js,jsx}", "**/__tests__/**/*.{ts,tsx,js,jsx}"],
       "env": {
         "node": true,
-        "jest": true
+        "jest": true,
+        "jest/globals": true
+      },
+      "plugins": ["jest", "no-snapshot-testing", "testing-library"],
+      "extends": [
```
esezen, line 69: I think we only need 
```
        "plugin:jest/recommended",
        "plugin:testing-library/react"
```
here as the rest of them already come from the main config

.eslintrc.json
```diff
@@ -52,22 +52,40 @@
...
+        "plugin:@cspell/recommended",
+        "plugin:jest/recommended",
+        "plugin:testing-library/react"
+      ],
+      "rules": {
+        // https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-node-access.md
+        "testing-library/no-node-access": ["error", { "allowContainerFirstChild": true }],
+        // https://github.com/VicJer/eslint-plugin-no-snapshot-testing
+        "no-snapshot-testing/no-snapshot-testing": "error",
+        "import/no-extraneous-dependencies": "off",
```
esezen, line 82: Can we scope this to `devDependencies`?

```
"import/no-extraneous-dependencies": [
      "error",
      {
        "devDependencies": [
          "**/stories/**/*.*",
          "**/.storybook/**/*.*",
          "spec/**/*.*"
        ],
        "peerDependencies": true
      }
    ]
```

.eslintrc.json
```diff
@@ -52,22 +52,40 @@
...
+        // https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-props-no-spreading.md
+        "react/jsx-props-no-spreading": "off",
+        "@cspell/spellchecker": ["error"],
```
esezen, line 85: Do we need this?

.eslintrc.json
```diff
@@ -52,22 +52,40 @@
...
+        "jest/valid-title": "off"
```
esezen, line 86: Do we need this?

.husky/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
```
esezen, line 1: @v-i-s-h-n-u-ps I have some opinions about this 😅 

Do you feel strongly about making this pre-commit or would you be okay moving this to a pre-push? IMHO, pre-commit hooks disrupt the developer's workflow and do more harm than good. I also don't love the fix tag, I think the automatic run should show you the error but shouldn't fix by itself. 

I'm happy to have a conversation about his if you feel strongly. My proposal is to make this a pre-push hook and remove the `--fix` (this part is less important). Here is an example from one of our other repos: https://github.com/Constructor-io/constructorio-client-javascript/blob/master/.husky/pre-push

esezen (commented):
@v-i-s-h-n-u-ps this is looking great. Thanks for making on these changes and making our library better! I left some comments here and there. They are mostly about the configurations and not the tests themselves, the tests are looking great. Let me know what you think and we can also hop on a call and take a look at them together

src/components/SkipButton/SkipButton.test.tsx
esezen: I saw this after mentioning it in my review. I kinda do have a strong preference just because we follow that pattern in other libraries and I would love to keep it consistent if possible

esezen: Side note about this, since tests files are under `src` our compile step is trying to build them and it's producing some errors. Depending on where we land with this decision, we might need to look into that

.eslintrc.json
esezen: But even in tests we don't need to import something that we didn't include in `package.json`, no? We want to use stuff from 'devDependencies' in tests but not something that we didn't include in any of the dependencies

esezen (approved):
LGTM!

PR #180, opened 2024-03-29
[QUIZ-723]: subtask tests for rest of the files

Gull-Dev:
Have added tests for hooks server & client and utils


v-i-s-h-n-u-ps (changes requested):
@Gull-Dev Thank you for covering more files with unit tests. A couple of things, Can we keep same file structure as we've in `src` for the tests. Also, `useQuizApiState.ts`, `useSelectInputProps.ts` can use more coverage.

esezen (approved):

v-i-s-h-n-u-ps (approved):

PR #182, opened 2024-04-18
added results

v-i-s-h-n-u-ps:


src/components/ShareResultsModal/ShareResultsModal.tsx
```diff
@@ -42,7 +43,9 @@ export default function ShareResultsModal({
               ? 'Share or save your quiz results through email or using the link below.'
               : 'Share or save your quiz results with this link.'}
           </div>
-          {onEmailResults && <EmailField onSubmit={(email) => onEmailResults({ email, url })} />}
+          {onEmailResults && (
+            <EmailField onSubmit={(email) => onEmailResults({ email, url, results })} />
```
esezen, line 47: The name results might be a bit confusing here since `results` is actually the `response` object that we get from the API which has `request` and `response` which has `results` under it 😅 
We can either rename this to something else and keep exposing the same value or alternatively we can expose `quizState.results.response.results`. I am okay with both. What do you think @v-i-s-h-n-u-ps ?
Screenshot 2024-04-19 at 2 15 05 PM

v-i-s-h-n-u-ps: It seems enough to expose `quizState.results.response.results`. Updated [9cffc91](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/182/commits/9cffc91c1f05da072ecd5db34d826d56afdc5961)

PR #192, opened 2024-07-31
Ci 3728 os UI quizzes support closing toast notifications on share

mocca102:
- Added a close icon for the toaster
- Refactored Toaster
- Added tests 
Screenshot 2024-07-31 at 2 58 47 PM

### DIFF
--- a/src/components/ShareResultsModal/EmailField.tsx
+++ b/src/components/ShareResultsModal/EmailField.tsx
@@ -1,7 +1,8 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import CheckMarkCircleSVG from './CheckMarkCircleSVG';
 import AlertCircleSVG from './AlertCircleSVG';
+import Toaster from './Toaster';
 
 interface EmailFieldProps {
   onSubmit: (email: string) => Promise<void>;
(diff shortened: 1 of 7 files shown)


src/components/ShareResultsModal/EmailField.tsx
```diff
@@ -42,6 +44,18 @@ export default function EmailField({ onSubmit }: EmailFieldProps) {
     }
   };
 
+  const onCloseToaster = () => {
+    setShowNotification(false);
+    setIsSubmitted(false);
+    setFormError(null);
+  };
+
+  useEffect(() => {
```
esezen, line 53: Instead of using a `useEffect` here, would it be better to `setShowNotification` in the same place we set `isSubmitted` and `formError`?

src/components/ShareResultsModal/LinkField.tsx
```diff
@@ -10,6 +11,19 @@ interface LinkFieldProps {
...
   const [isError, setIsError] = useState(false);
+  const [showNotification, setShowNotification] = React.useState(false);
+
+  const onCloseToaster = () => {
+    setShowNotification(false);
+    setIsCopied(false);
+    setIsError(false);
+  };
+
+  useEffect(() => {
```
esezen, line 22: Instead of using a `useEffect` here, would it be better to `setShowNotification` in the same place we set `isSubmitted` and `formError`?

mocca102: I get that it's simpler this way and more direct. But I like to think of the notification as a side-effect of submitting the form not as part of the `handleSubmit` logic itself. (Not applicable here) but say for example some other function than `handleSubmit` is updating `isSubmitted` or `formError`, we wouldn't worry about adding `setShowNotification` there as well. All we will need is the `useEffect` that run whenever  `isSubmitted` or `formError` despite of how they were changed

esezen (approved):

PR #205, opened 2024-10-11
update email error logic

stanlp1:
Adds a disabled class while awaiting email submit callback.

### DIFF
--- a/src/components/ShareResultsModal/EmailField.tsx
+++ b/src/components/ShareResultsModal/EmailField.tsx
@@ -16,6 +16,7 @@ interface FormError {
 export default function EmailField({ onSubmit }: EmailFieldProps) {
   const [formError, setFormError] = useState<FormError | null>(null);
   const [isSubmitted, setIsSubmitted] = useState(false);
+  const [isInProgress, setIsInProgress] = useState(false);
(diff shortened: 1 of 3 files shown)


src/components/ShareResultsModal/EmailField.tsx
```diff
@@ -30,10 +31,16 @@ export default function EmailField({ onSubmit }: EmailFieldProps) {
 
   const handleSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
     e.preventDefault();
```
mocca102, line 33:
```suggestion
    setIsSubmitted(false);
    e.preventDefault();
```

mocca102: Currently 

1. If you click on "Send" for the first time you get the success notification
2. If you do not close the notification and click again on "Send" nothing happen (The notification message doesn't show again) (It does for the error message though)

I believe this is because "isSubmitted" doesn't get updated. The suggested line of code should fix the issue

stanlp1: 👍 Yep made that change.

mocca102 (approved):
LGTM!

PR #216, opened 2024-12-24
[QUIZ-798] quizzes UI update share results link

v-i-s-h-n-u-ps:


src/hooks/useQuizState/useQuizApiState.ts
```diff
@@ -103,8 +120,13 @@ const useQuizApiState: UseQuizApiState = (
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [quizApiState.resultsConfig]);
 
+  const prevQuizId = usePrevious(quizId);
+
   useEffect(() => {
     (async () => {
+      // If quizId is the same as the previous quizId, wait for next render after reset
+      if (!!prevQuizId && quizId !== prevQuizId) return;
```
jjl014, line 128: Would you be able to elaborate a bit more on what this is for and what it's doing? Thanks!

jjl014 (commented):
This is looking pretty good from what I can tell! 🚀  We'll want someone from the CI team that's more familiar with the library to provide a more thorough review though.

I was also curious about what `usePrevious` is doing if you could provide a bit more context there. Thanks!

src/hooks/useQuizState/useQuizApiState.ts
mocca102: +1 on Jimmy's ask. Why are we keeping track of a `prevQuizId` state?

mocca102 (changes requested):
Thank you for working on this @v-i-s-h-n-u-ps 

Looking mostly good but I don't think we need the usePrevious logic. Please correct me if I'm wrong

src/hooks/useQuizState/useQuizApiState.ts
```diff
@@ -85,7 +92,17 @@ const useQuizApiState: UseQuizApiState = (
...
-      const quizResults = await getBrowseResultsForItemIds(cioClient, queryItems);
+      let quizResults;
+      try {
+        quizResults = await getQuizResults(cioClient, quizId, {
+          answers,
+          resultsPerPage: resultsPageOptions?.numResultsToDisplay,
+          quizVersionId: quizVersionIdFromParam,
+          ...resultsPageOptions?.requestConfigs,
+        });
+      } catch (error) {
```
mocca102, line 103: Can you please explain why this catch is being in place? Just want to understand when do you expect the getQuizResults to fail

mocca102: @v-i-s-h-n-u-ps

v-i-s-h-n-u-ps: In case the version of the quiz is no longer available. We fall back to the original behaviour in that case.

src/types.ts
```diff
@@ -123,6 +123,7 @@ export interface IQuizProps {
 export interface QuizReturnState {
   answers: {
     inputs: AnswerInputState; // Key is the question Id and value is the answer input
+    answers: string[][];
```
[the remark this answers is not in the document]
v-i-s-h-n-u-ps: [207f8fe](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/216/commits/207f8fe483a6b125eb22caea2cf67a047b597137)

src/hooks/useQuizState/useQuizApiState.ts
v-i-s-h-n-u-ps: There was an issue with merging my remote branches. Removed here [207f8fe](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/216/commits/207f8fe483a6b125eb22caea2cf67a047b597137)

mocca102 (approved):
LGTM!

PR #223, opened 2025-02-12
Support getResultCardImage url for a custom image url

mocca102:
Support passing a custom image url for result card

This will allow customers to tweak their image urls to add `width` and `height` queries for example to their images to support different image sizes

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -53,6 +53,7 @@ export interface ResultCardOptions {
     },
     index: number
   ) => JSX.Element;
+  getResultCardImageUrl?: (imageUrl: string) => string;
 }
 
 export namespace QuizResultsEventsProps {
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -53,6 +53,7 @@ export interface ResultCardOptions {
     },
     index: number
   ) => JSX.Element;
+  getResultCardImageUrl?: (imageUrl: string) => string;
```
stanlp1, line 56: Should we pass the item data as well? 

In case some customers have logic that depends on some metadata.

stanlp1 (approved):
LGTM. One comment.

src/types.ts
mocca102: Great callout

PR #226, opened 2025-03-07
add callback for open question

Gull-Dev:
Update the UI library with to provide additional callback for the merchandiser

### DIFF
--- a/src/hooks/usePropsGetters/useNextQuestionButtonProps.ts
+++ b/src/hooks/usePropsGetters/useNextQuestionButtonProps.ts
@@ -7,7 +7,7 @@ export default function useNextQuestionButtonProps(
   nextQuestion: QuizEventsReturn.NextQuestion,
   quizApiState: QuizAPIReducerState,
   quizLocalState: QuizLocalReducerState,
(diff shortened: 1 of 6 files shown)


src/hooks/usePropsGetters/useNextQuestionButtonProps.ts
```diff
@@ -24,10 +27,22 @@ export default function useNextQuestionButtonProps(
       'aria-describedby': buttonDisabled ? 'next-button-help' : '',
       type: 'button',
       onClick: () => {
+        if (currentQuestion?.type === 'open' && onOpenQuestionAnswered && currentQuestionId) {
+          const value = quizLocalState.answerInputs[currentQuestionId]?.value;
+
+          if (value && typeof value === 'string') {
+            onOpenQuestionAnswered(value, currentQuestion);
```
mocca102, line 34: @Gull-Dev I believe this is already supported on `onQuizNextQuestion` callback inside `callbacks` passed to `IQuizProps`

And the callback takes these two parameters same as your example `currentQuestion, answer: currentAnswerInput`

check the call of `onQuizNextQuestion` inside `useQuizNextClick` hook

PR #292, opened 2026-05-07
At 100/quiz option description websitejs libraryui library

dmitrycnstrc:
Add support for displaying option-level descriptions in single-select quiz questions
When a single option with a description is selected, the question description area shows the option's description instead of the question-level description
Disable auto-advance to next question when the selected option has a description (user must click Continue)

Test plan
- Single-select question: click option with description → description text updates to option's description
- Single-select question: click option without description → shows original question description
- Single-select question with description option: auto-advance does NOT trigger, user clicks Continue
- Single-select question without description options: auto-advance works as before
- Multiple-select questions: no behavior change
- Cover/OpenText questions: no behavior change

### DIFF
--- a/spec/utils/utils.test.ts
+++ b/spec/utils/utils.test.ts
@@ -14,6 +14,7 @@ import {
   getDisplayedDescription,
 } from '../../src/utils';
 import { QuestionTypes } from '../../src/components/CioQuiz/actions';
+import { Question } from '../../src/types';
 
 describe('convertPrimaryColorsToString', () => {
   it('returns empty string when no images are provided', () => {
@@ -288,8 +289,11 @@ describe('resetQuizSessionStorageState', () => {
 });
 
 describe('getDisplayedDescription', () => {
-  const makeQuestion = (type: string, description?: string, options?: any[]) =>
-    ({ type, description, options } as any);
+  const makeQuestion = (
+    type: string,
+    description?: string,
+    options?: { id: number | string; value: string; description?: string }[]
+  ) => ({ type, description, options } as unknown as Question);
 
   it('returns undefined when question is null or undefined', () => {
     expect(getDisplayedDescription(null, {})).toBeUndefined();
--- a/spec/utils/utils.test.ts
+++ b/spec/utils/utils.test.ts
@@ -11,6 +11,7 @@ import {
   getStateFromSessionStorage,
   isFunction,
   convertPrimaryColorsToString,
+  getDisplayedDescription,
 } from '../../src/utils';
 import { QuestionTypes } from '../../src/components/CioQuiz/actions';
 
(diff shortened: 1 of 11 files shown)


dmitrycnstrc:
@claude

src/components/SelectTypeQuestion/SelectTypeQuestion.tsx
```diff
@@ -3,16 +3,18 @@ import { Nullable } from '@constructor-io/constructorio-client-javascript/lib/ty
...
-import { Question, QuestionOption } from '../../types';
-import { renderImages } from '../../utils';
+import { Question, QuestionOption, Selected } from '../../types';
+import { getDisplayedDescription, renderImages } from '../../utils';
 import { QuestionTypes } from '../CioQuiz/actions';
 
-export interface Selected {
-  [key: number]: boolean;
-}
+export type { Selected };
```
Alexey-Pavlov: Is this re-export really necessary?

src/hooks/usePropsGetters/useSelectInputProps.ts
Alexey-Pavlov: Sorry, it's pre-existing, but it looks like this ref should be a boolean, not an object; it's later assigned `true` and checked as a boolean.

```suggestion
  const singleSelectClicked = useRef(false);
```

src/utils.tsx
```diff
@@ -247,3 +247,25 @@ export function getNestedValueUsingDotNotation(object: any, key?: string): unkno
 
   return key.split('.').reduce((a, b) => a?.[b], object);
 }
+
+export function getDisplayedDescription(
+  question: Question | null | undefined,
+  selected: Selected
+): string | undefined {
+  if (!question?.description && !question?.options) return undefined;
```
Alexey-Pavlov:
```suggestion
  if (!question) return undefined;
```

What if we use a simpler guard here? Downstream logic handles all the nullable paths already

spec/utils/utils.test.ts
```diff
@@ -285,3 +286,61 @@ describe('resetQuizSessionStorageState', () => {
     );
   });
 });
+
+describe('getDisplayedDescription', () => {
+  const makeQuestion = (type: string, description?: string, options?: any[]) =>
+    ({ type, description, options } as any);
```
Alexey-Pavlov: How about using `Question` type here? 
Something like 
```ts
const makeQuestion = (
    type: string,
    description?: string,
    options?: { id: number | string; value: string; description?: string }[]
  ) => ({ type, description, options } as unknown as Question);
```

Alexey-Pavlov (changes requested):
Thank you for working on this! Left just a few comments

src/components/SelectTypeQuestion/SelectTypeQuestion.tsx
dmitrycnstrc: [bf1c800](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/292/commits/bf1c800dad5bf9e8474dd5eca4de67dcf356a658)

src/hooks/usePropsGetters/useSelectInputProps.ts
dmitrycnstrc: [5e5668c](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/292/commits/5e5668c19affae639c8c3d8ec454f297b80bd069)

src/utils.tsx
dmitrycnstrc: [9052bb7](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/292/commits/9052bb7c288d529b3ba1c6850c26c52473943947)

spec/utils/utils.test.ts
dmitrycnstrc: [a98c783](https://github.com/Constructor-io/constructorio-ui-quizzes/pull/292/commits/a98c783e67fda1fc160cca2b32d0822ce7c31ba6)

Alexey-Pavlov (approved):
LGTM! Thank you!

PR #296, opened 2026-05-11
[AT-82]: quizzes asa message

dmitrycnstrc:
**Summary**
* Display `quiz_asa_results_message` from the quiz results API response instead of the default response summary ("Based on your answers...") on the results page
* Falls back to the existing matched-options summary when the field is null or absent
* Added Storybook story to preview the ASA message behavior

**Test plan**
*  Verify ASA message renders when quiz_asa_results_message is present in the API response
*  Verify default response summary renders when quiz_asa_results_message is null
*  Verify default response summary renders when quiz_asa_results_message is absent from the response
*  Check Storybook story QuizResultsPageWithAsaMessage displays the ASA message correctly

Here are your results

### DIFF
--- a/src/components/ResultFiltersAndShare/ResultFiltersAndShare.tsx
+++ b/src/components/ResultFiltersAndShare/ResultFiltersAndShare.tsx
@@ -2,6 +2,7 @@ import React, { useContext } from 'react';
 import QuizContext from '../CioQuiz/context';
 import ShareButton from '../ShareButton/ShareButton';
 import QuizResultsSummary from '../QuizResultsSummary/QuizResultsSummary';
+import { QuizResultsResponse } from '../../types';
 
 interface ResultFiltersAndShareProps {
   onShare: () => void;
@@ -17,11 +18,13 @@ function ResultFiltersAndShare({
   const { state } = useContext(QuizContext);
   const matchedOptions = state?.quiz?.matchedOptions;
   const summary = state?.quiz?.resultsConfig?.desktop?.response_summary ?? null;
+  const asaMessage =
+    (state?.quiz?.results as QuizResultsResponse | undefined)?.quiz_asa_results_message ?? null;
 
   return (
     <div className='cio-results-filter-and-redo-container cio-results-button-group'>
       <div className='cio-results-filter-container'>
-        <QuizResultsSummary summary={summary} matchedOptions={matchedOptions} />
+        <QuizResultsSummary summary={summary} matchedOptions={matchedOptions} asaMessage={asaMessage} />
       </div>
       <div className='cio-results-number-and-share-button-group'>
         {numberOfResults} {numberOfResults === 1 ? 'result' : 'results'}
--- a/src/components/ResultFiltersAndShare/ResultFiltersAndShare.tsx
+++ b/src/components/ResultFiltersAndShare/ResultFiltersAndShare.tsx
@@ -25,7 +25,11 @@ function ResultFiltersAndShare({
   const matchedOptions = state?.quiz?.matchedOptions;
   const summary = state?.quiz?.resultsConfig?.desktop?.response_summary ?? null;
   const results = state?.quiz?.results;
-  const asaMessage = (isQuizResultsResponse(results) && results.quiz_asa_results_message) || null;
+  const asaMessage =
+    (isQuizResultsResponse(results) &&
+      typeof results.quiz_asa_results_message === 'string' &&
+      results.quiz_asa_results_message) ||
+    null;
 
   return (
     <div className='cio-results-filter-and-redo-container cio-results-button-group'>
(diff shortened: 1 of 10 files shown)


src/components/ResultFiltersAndShare/ResultFiltersAndShare.tsx
```diff
@@ -17,11 +18,17 @@ function ResultFiltersAndShare({
   const { state } = useContext(QuizContext);
   const matchedOptions = state?.quiz?.matchedOptions;
   const summary = state?.quiz?.resultsConfig?.desktop?.response_summary ?? null;
+  const asaMessage =
+    (state?.quiz?.results as QuizResultsResponse | undefined)?.quiz_asa_results_message ?? null;
```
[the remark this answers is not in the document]
dmitrycnstrc: d29fc43

src/components/ResultFiltersAndShare/ResultFiltersAndShare.tsx
```diff
@@ -2,6 +2,13 @@ import React, { useContext } from 'react';
 import QuizContext from '../CioQuiz/context';
 import ShareButton from '../ShareButton/ShareButton';
 import QuizResultsSummary from '../QuizResultsSummary/QuizResultsSummary';
+import { QuizResultsResponse, QuizSharedResultsData } from '../../types';
+
+function isQuizResultsResponse(
+  results: QuizResultsResponse | QuizSharedResultsData | undefined
+): results is QuizResultsResponse {
+  return !!results && 'quiz_id' in results;
+}
```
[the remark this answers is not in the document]
dmitrycnstrc: 8aa5ce0

src/components/ResultFiltersAndShare/ResultFiltersAndShare.tsx
```diff
@@ -17,11 +24,17 @@ function ResultFiltersAndShare({
   const { state } = useContext(QuizContext);
   const matchedOptions = state?.quiz?.matchedOptions;
   const summary = state?.quiz?.resultsConfig?.desktop?.response_summary ?? null;
+  const results = state?.quiz?.results;
+  const asaMessage = (isQuizResultsResponse(results) && results.quiz_asa_results_message) || null;
```
esezen: Does `quiz_asa_results_message` under the response type? AFAIK, it doesn't and we need to update client-js.

esezen: Also why are we checking for `isQuisResultsResponse`?

esezen: Last but not least, should we also check if this field is string rather than just truthy + non empty?

src/stories/Quiz/tests/mocks.tsx
```diff
@@ -19,7 +19,8 @@ import useCoverQuestionProps from '../../../hooks/usePropsGetters/useCoverQuesti
 import useSelectInputProps from '../../../hooks/usePropsGetters/useSelectInputProps';
 
 type MockOptions = {
-  withSummary: boolean;
+  withSummary?: boolean;
+  asaMessage?: string | null;
```
esezen: Why do we need `null` here? Would we ever explicitly pass `null` instead of not passing it?

src/components/ResultFiltersAndShare/ResultFiltersAndShare.tsx
esezen: How does `quiz_asa_results_message` gets populated? We need to make sure there is no surprises for the consumers. If this can populate without them taking explicit action, we have to put this behind a feature flag

esezen (changes requested):

src/components/ResultFiltersAndShare/ResultFiltersAndShare.tsx
dmitrycnstrc: In the dashboard, a customer can add a question with the free_form type to their quiz. When such a question exists, the quiz-service calls ASA to fetch results and the accompanying description, which gets passed to the client as quiz_asa_results_message
This field won't appear for consumers unless they explicitly add an free_form type question to their quiz — it's an opt-in action on the dashboard side, not something that shows up unexpectedly

src/stories/Quiz/tests/mocks.tsx
dmitrycnstrc: Removed null 👍

src/components/ResultFiltersAndShare/ResultFiltersAndShare.tsx
dmitrycnstrc: Added check 👍

dmitrycnstrc: The check is needed because state.quiz.results can be either QuizResultsResponse or QuizSharedResultsData. quiz_asa_results_message only exists on QuizResultsResponse, so the guard ensures we don't access it on the wrong response type

dmitrycnstrc: good catch, forgot about it, thank you! i've created PR, could you take a look please? 
https://github.com/Constructor-io/constructorio-client-javascript/pull/458

esezen (approved):

